PR #2, opened 2020-10-19
Main nav

brown-a2:
Main menu changes include justifying the menu so that it aligns flush with the container edge per Toni's designs. Also, this PR moves the logo inside the container so everything is aligned. However, further work needs to be done to align our blocks so that they sit within the container.


BNewing (approved):
Looks good to me :)

PR #3, opened 2020-10-20
Default template, page header, header colour, light teal

RobLoweMOJ:
Added page header (with intro excerpt) to default template
Full width now uses a different template to default.
Added customizer option for page header colour option
Reworded breadcrumb
Added Light teal.


brown-a2 (approved):

assets/scss/page.scss
```diff
@@ -0,0 +1,45 @@
...
+    right: 50%;
+    margin-left: -50vw;
+    margin-right: -50vw;
+    background: rgba(#000000, 0.1);
+    padding: 1em 0 0 0;
+    margin-bottom: 2em;
+    display: inline-block;
+
+    @include mq($from: tablet) {
+      padding: 2em 0 0 0;
```
BNewing, line 18: I think it's fine for now, but we probably want to add in px fallbacks where we can :) it looks like vh support is a bit limited too but I'll take a look when i use it :)

BNewing (approved):

PR #5, opened 2020-10-20
Unstyled base for search template

BNewing:
This PR adds in all of the base functionality required in the search template:

- [x] Full width search items
- [x] Search bar included in page template
- [x] Number of results is at the top of the page
- [x] Date last updated is included in search item preview
- [x] Pagination is text that says how many pages there are instead of links

I will work on styling it to the design separately, this is just to get it up on the site so anyone else can use it and see it if they need it.


brown-a2 (approved):

assets/scss/header.scss
```diff
@@ -311,4 +311,40 @@
...
+	margin-left: 0;
+
+	& .nhsuk-search__input {
+		float: none;
+		width: calc(100% - 54px);
+		display: inline-block;
+	}
+
+	& .nhsuk-search__submit {
+		float: none !important;
```
RobLoweMOJ, line 326: Just double checking theres no way we can use class names (or ids) rather then important

BNewing: Yes using important is a bit of a last resort. I went with it to get something done quick, and to override something else that was quite high specificity, but I'll take another look at removing it with the next round of work on this where I focus more on the styling :)

PR #7, opened 2020-10-20
Search styling for desktop

BNewing:
This adds in Toni's designs to the search page, for desktop.

Mobile, tablet and hover/focus states will be done in a separate PR. This is to share the work so that others can have a look at it and use it too.

### DIFF
--- a/assets/scss/header.scss
+++ b/assets/scss/header.scss
@@ -316,16 +316,24 @@
 	float: unset;
 	margin-left: 0;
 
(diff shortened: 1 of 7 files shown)


RobLoweMOJ (approved):
Looks good only thing i would suggest is to move the colours in search.scss to variables. they are currently listed out in nhs-colors.scss.

brown-a2 (approved):

BNewing:
Thank you both - will do @RobLoweMOJ!

PR #11, opened 2020-10-22
Search page fine tuning

BNewing:
Fixes bugs found with the search, after the base desktop design was done:

- [x] Pagination button text needed visually to hide on mobile (but still be present for screen reader users)
- [x] Search bar had a dark blue bar under it
- [x] Search button was resizing oddly on mobile view
- [x] Additional, unnecessary search buttons were appearing on the search on mobile view
- [x] Shifts some of the colours to use variables instead of hex codes
- [x] Focus, hover, active states on interactive elements on the page


RobLoweMOJ (approved):
Looks good

PR #13, opened 2020-10-22
Layout alignment fix

BNewing:
There was an issue where the header text wasn't aligned with the body text on all screen widths. This pulls it into alignment.


RobLoweMOJ (approved):
Looks good

PR #14, opened 2020-10-22
Bug squash

brown-a2:
Stops page background colour from appearing where it shouldn't :< Ie looped pages where the page id is used such as search.


brown-a2:
Could also use the is_singular() function here but it seems is_page() will suffice for now.

PR #19, opened 2020-10-23
Add in ARIA to toggle button text between close and open

BNewing:
When you open the menu button, the open button becomes the close button. This was conveyed visually with the close icon, but there wasn't anything for screen reader users.

This adds in text that a screen reader user will here, that switches between 'close menu' and 'open menu', so DAC can test on Monday

### DIFF
--- a/assets/scss/header.scss
+++ b/assets/scss/header.scss
@@ -19,6 +19,11 @@
     }
   }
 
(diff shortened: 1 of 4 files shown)


brown-a2 (approved):

RobLoweMOJ (approved):
👍

PR #20, opened 2020-10-23
Remove yellow focus states

BNewing:
There were a couple of yellow focus states added in, which I've gone through and removed: 

- [x] Links in body text
- [x] Links in footer text
- [x] Breadcrumb links

There were a couple of other things too that this addresses:
- [x] Multiple home links in the header when we just need one
- [x] focus state on home link

I'm sure it could have been done in a neater way, but I was keen to get a PR done in case we can merge it before DAC testing, as they're things that they'll probably pick up on.

There's another task for updating the card focus links to match, but that's in the MOJ blocks repo rather than here.


brown-a2 (approved):
Not sure if you noticed but when you click down on a menu item, you get a black underline effect. Does this fix that?

BNewing:
If it's what I think you mean, it doesn't - I did spot what was causing it, and got it to stop doing it on hover, but it's now doing it on focus instead. I spent a while trying to get it fixed and then decided it was maybe enough for now, given the yellow has gone, but it is bugging me...

brown-a2:
Oh okay cool, yes I think we're talking about the same thing. I was just wondering if I missed it, but it can be done in a separate PR :)

BNewing:
I'll put a note in that spreadsheet :D

PR #25, opened 2020-11-04
Update hover, focus and active states to match design

BNewing:
The active state didn't match the designs. There was also an issue where when you clicked on an icon, the hover state overrode the active state and caused the text to disappear. This fixes both of those issues.


RobLoweMOJ (approved):
👍

PR #26, opened 2020-11-05
Add sidebar metabox selector to turn sidebar display on page on/off

brown-a2:
- Added metabox on page to allow an editor to turn sidebar on or off on the page. Default is sidebar is displayed.
- I've tried to leave in place some of the sidebar logic as we may still be using widgets with the sidebar. We can remove some of this later if it turns out we don't want to use widgets via Customiser in the sidebar.
- I've also rationalized all the page metaboxes so that they are now organized similar to each other. However, we may want to take this a step and lump them all under one title. To be discussed further.
- Formatted some of the PHP -> php12

I've left the `full-width` template in this commit. I will remove in a following commit when I've changed the pages on production over to the default template.


RobLoweMOJ (commented):
Just checking do we want to remove ability to remove the page title? I guess at the moment its only the home pages but should it be available on other pages?

brown-a2:
This should have been in the PR but I thought it was going (except on the homepage)? It seems like a bad situation to have the editor be able to create a page without a title, but open to suggestions.

RobLoweMOJ (approved):
👍

PR #30, opened 2020-11-06
Remove Critical CSS, NHS Logo and sitemap

RobLoweMOJ:


BNewing (approved):
Looks good to me :) I quite like the idea of critical CSS, but I don't think what we had before now was really working for us (and I don't think making it work for us is a priority atm, maybe something to revisit in the future)

PR #31, opened 2020-11-11
Refactor default page and fix page gap bug

brown-a2:
Using the default template on the homepage results in a gap between the
header of the page and the block being used. This fix removes the gap
and puts the extra spacing on only non-homepage headers.

I also refactored page.php so that it is a bit more clear how the
special category components load in.


RobLoweMOJ (approved):
👍

BNewing:
Had a look through, looks good! Two thoughts - could the pagination maybe a list? Looking at GOV.UK, that's how they do it: https://www.gov.uk/access-to-elected-office-fund/what-youll-get

The other thing was maybe using px fallbacks for rem  :) but they're both minor things, it looks great!

brown-a2:
I don't see why not, would pagination help on the accessibility front? Ie, is that why you think they might have done that? Yes, the pixel fallback is a good point. I'll add that in, in another PR. While I was doing this commit, it did occur to me that when we have chosen a new framework, it would help with some of these things.

BNewing:
If it's a list, a screen reader user would hear something like 'next, list item one of two' which is a bit of additional context and might help with navigation. More of a minor point though I think, more of an enhancement!

PR #34, opened 2020-11-12
Remove full width template

brown-a2:
Full width template was almost identical to the default template, so all the full width template functionality has now been moved into the default template and is running on production. This is the final step to remove the now unused full width template.

Also fixed a header order fix h2 -> h1


style.css
```diff
@@ -2,7 +2,7 @@
 /*
 Theme Name: Hale
 Text Domain: hale
-Version: 1.0.3
+Version: 1.0.2
```
RobLoweMOJ, line 5: This should be 1.0.3?

RobLoweMOJ (commented):
Just a small note about the version number. You may need to also check MagiRec when you push up as the homepage will be using the Full Width Template so you will need to set this up to use the new template

style.css
brown-a2: Thanks actually, it needs to be one higher, updated.

RobLoweMOJ (approved):

PR #36, opened 2020-11-16
Add in flex for sticky footer

BNewing:
This adds in some flex CSS, for browsers that can support it. It makes the footer stick to the bottom of the page, when there's not enough content to push it down. It hooks onto the footer element and the diev with an id of 'content', which are both put in there by the hale theme itself.


brown-a2:
Looks good to me.

PR #38, opened 2020-11-19
Add proper focus state to search page search button and bring in line

BNewing:
The search bar button on the search bar didn't have the right focus states. This brings that in. It also shunts the bar to the left slightly, so it's in line with the rest of the page content.


brown-a2 (commented):
Is the version number now going in reverse?

brown-a2 (approved):

BNewing:
Versioning fixed! Not sure what happened there. Thank you!

PR #40, opened 2020-12-30
Make sticky footer height vh not % to avoid background colour bug

BNewing:
The `height: 100%` was causing issues for the background colour on pages with the minty background. The colour was taking browser height, not content height. `min-height` doesn't work in our use case with %, but it does with `vh`. This unit is supported at a similar rate to flex, and this is wrappered in a `@supports` for flex.


RobLoweMOJ (approved):
👍

PR #42, opened 2020-12-31
Make banner full width above 1024px (to prevent bug)

BNewing:
This re-aligns the text in the call to action section in the header. 

I think we will need to add in another breakpoint in the future. The tablet variable is around 40rem (~640px) and the desktop variable is around 48rem (768px). This might be part of the discussions we have around frontend frameworks (we still need to choose one for Hale).


RobLoweMOJ (approved):
👍

PR #45, opened 2021-01-05
Change Compile Method to Webpack

RobLoweMOJ:
Moves Hale theme to use Webpack for asset compiling

Future improvement tasks

Sort Sass into folders
Combine JS files.


brown-a2 (approved):
Should we add the rtl.css file out of root as well? And remove the gulp file in the root?

RobLoweMOJ:
rtl.css and grunt root file removed.

PR #46, opened 2021-01-07
Add meta description tag, that links to the site tagline

BNewing:
Because the cookie banner comes naer the top of the DOM (for accessibility reasons), that text is currently being pulled out by google and shown in search results. This adds in a meta tag that pulls in the site tagline as the site description. This means this should show on google instead of the cookie info.

I've also updated the meta tag with the theme name in to say hale, instead of nightingale.

This is after discussion with Anna and Josh (Content Designers) to double check it's okay to use the tagline in this way, from their perspective.


RobLoweMOJ (approved):
👍

PR #47, opened 2021-02-17
New Page Options - Display breadcrumb , Display Title Section

RobLoweMOJ:


inc/breadcrumbs.php
```diff
@@ -80,11 +80,26 @@ function nightingale_uncanny_breadcrumb_check() {
 function nightingale_breadcrumb() {
```
brown-a2, line 80: Would it be possible to change this function name to Hale without too much difficulty? Just so that we are slowly refactoring as we go with the theme.

inc/breadcrumbs.php
```diff
@@ -80,11 +80,26 @@ function nightingale_uncanny_breadcrumb_check() {
...
+
+        if(is_singular('page')) {
+            global $post;
+
+            // Get breadcrumb setting, is it set to display or not
+            $show_page_breadcrumb_setting = get_post_meta($post->ID, 'hale_metabox_page_breadcrumb', true);
+
+            if(!empty($show_page_breadcrumb_setting)){
+                $show_page_breadcrumb = $show_page_breadcrumb_setting;
+            }
```
brown-a2, line 97: Just checking, will the `get_post_meta($post->ID, 'hale_metabox_page_breadcrumb', true);` return a "yes" and also do you need the check if it's empty if you are just changing one variable for another but not setting anything new?

brown-a2 (approved):

inc/breadcrumbs.php
RobLoweMOJ: Done

inc/breadcrumbs.php
RobLoweMOJ: it would return a "yes" or "no" if set or if its not set empty. theres 2 variables  $show_page_breadcrumb which defaults as 'yes' unless $show_page_breadcrumb_setting is set.

PR #52, opened 2021-03-02
Fix typo causing PHP index error

brown-a2:
* Minor fix to a PHP error being thrown, caused by the wrong variable being set.

### DIFF
--- a/inc/template-functions.php
+++ b/inc/template-functions.php
@@ -148,14 +148,14 @@ function nightingale_sidebar_location( $sidebar ) {
  */
 function nightingale_custom_page_colour( $classes ) {
 
-    $colour_array      = array(
+    $colour_array = [
         '005eb8' => 'blue',
         '0c223f' => 'dark-blue',
         '336c83' => 'teal',
         '00a19a' => 'light-teal',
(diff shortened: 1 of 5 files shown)


inc/template-functions.php
```diff
@@ -166,7 +166,7 @@ function nightingale_custom_page_colour( $classes ) {
     $page_header_colour = get_theme_mod( 'page_header_colour', '' );
 
     if(!empty($page_header_colour)){
-        $page_header_colour_name = 'page-header-color--' . $colour_array[ $page_header_colour ];
+        $page_header_colour_name = 'page-header-colour--' . $colour_array[ $page_header_colour ];
```
RobLoweMOJ, line 169: if we changing color to colour we also need to do it in the sass. (page-colours.scss)

brown-a2: Changed. Hopefully I didn't miss any of them. I've left the American spelling in the main colours.scss for now as Malcolm is working on that ticket.

brown-a2: I missed one, however now fixed. Grep is saying there are no more :|

PR #54, opened 2021-03-10
Colours

MalcolmVonMoJ:
- Changed `page-colours.scss` to declare colours used in each branding.
- Also in `page-colours.scss`, function to apply colours to brands.
- New file (`brandings.scss`) to apply above function and style all brands.
- Added specific branding styles in new file (`specific-brandings.scss`).
- Old `colours.scss` and `theme-colours.scss` file deleted.
- Made the new "neptune" brand the default in `template-functions.php`.
- Deleted unused PHP code.
- Rationalised WP options by removing secondary colour selection and renaming the colour picker to "branding".


assets/scss/colours.scss
```diff
@@ -1,28 +0,0 @@
...
-$color_lavendar-purple: #7E57C2;
-$color_yellow: #ffeb3b;
-$color_sunflower-yellow: #FFE600;
-$color_warm-yellow: #FFB81C;
-$color_red: #da291c;
-$color_teal: #336C83;
-$color_light-teal: #00a19a;
-$color_aqua-green: #00a499;
-$color_dark-green: #006747;
-$color_green: #007F3B;
```
RobLoweMOJ, line 28: Just wanted check we dont need some of these colours as bluebell was being used for the default link colour  and lavendar for visited link

RobLoweMOJ (commented):
Looks good. Left a comment.

assets/scss/footer.scss
```diff
@@ -1,23 +1,21 @@
-.hale-page {
-  .govuk-footer .widget_social_widget {
-    //the styles here are just for the social media icon links.
-    font-size: 40px;
+.govuk-footer .widget_social_widget {
+  //the styles here are just for the social media icon links.
+  font-size: 40px;
```
BNewing, line 3: Could we add in rem as well, so we can do the browser-based font enlarge?

assets/scss/page-colours.scss
```diff
@@ -1,65 +1,141 @@
...
-
-@each $name,
-$colour in $page-colours {
-
-  .page-header-colour--#{$name} {
-
-    .page-header {
-      background: rgba($colour, 0.1);
+//Uranis (Magistrates) branding
+$hale-uranus-brand-colour: #336C83;
```
BNewing, line 8: Hehe where did uranus come from?

BNewing (commented):
Looks good! Have we gone for a planetary naming scheme? What's HMG - I saw that in amongst them at one point :)

The only thing I'm not sure about is about ems/rem/px - it'd be nice if we could make the theme rem/px based, so we can do that browser resizing thing. It's not something that we'd pull out into a separate ticket, but we've been trying to work it in on PRs we can. I saw you put some ems in - am I right in thinking they're better supported than rem? I normally go rem because I'm nervous of the compounding!

Otherwise though it looks great - great work :)

assets/scss/page-colours.scss
MalcolmVonMoJ: Uranus is teal coloured...

assets/scss/footer.scss
MalcolmVonMoJ: Changed.

assets/scss/colours.scss
MalcolmVonMoJ: Where these were needed for link colours etc, they've been re-assigned elsewhere.  But I have retained these colours in a comment in `page-colours.scss` just in case we want to know what they were.

PR #55, opened 2021-03-12
JavaScript for new framework

MalcolmVonMoJ:
- Adds in new GDS JavaScript.
- Adds new custom JS for searchbox show-hide.
- Related CSS and HTML changes needed for the GDS JS and for the search show-hide.

### DIFF
--- a/assets/js/hale-scripts.js
+++ b/assets/js/hale-scripts.js
@@ -1,0 +1,14 @@
+/*
+	hale script v1.0.0
+*/
+
+/*
+	Search button in header - show-hide the search field
+*/
+document.getElementById("search-show-hide").addEventListener('click', 
+	function(){
+		var ariaExpandedStatus = this.getAttribute("aria-expanded");
+		ariaExpandedStatus = ariaExpandedStatus == "true" ? "false" : "true";
+		this.setAttribute("aria-expanded", ariaExpandedStatus);
+	}
+);
--- a/header.php
+++ b/header.php
@@ -25,7 +25,11 @@
 	flush();
 	?>
 </head>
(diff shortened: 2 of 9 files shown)


header.php
```diff
@@ -25,7 +25,11 @@
 	flush();
 	?>
 </head>
-<body <?php body_class( 'js-enabled' ); ?>>
+<body <?php body_class(""); ?>>
+<script>
```
BNewing, line 29: What was the thinking behind putting it in a script tag vs in the JS file?

assets/js/hale-scripts.js
```diff
@@ -0,0 +1,14 @@
+/*
+	hale script v1.0.0
+*/
+
+/*
+	Search button in header - show-hide the search field
+*/
+document.getElementById("search-show-hide").addEventListener('click',
```
BNewing, line 8: Do we need one for the hamburger menu too? (Sorry if that's been done in another PR!)

BNewing:
Looks good! Left two questions but I suspect they're probably done in other branches so I'll take a look at other PRs now :)

assets/js/hale-scripts.js
MalcolmVonMoJ: The menu one is dealt with by the GDS JavaScript already.

header.php
MalcolmVonMoJ: This the the approach taken by GDS - I have not investigated why they choose this method.  However, you can see it is the way they tell people to do it [in their get started guide](https://frontend.design-system.service.gov.uk/get-started/#5-get-the-javascript-working).

PR #57, opened 2021-04-13
Stop MoJ Blocks plugin from loading its own GovUK JS

brown-a2:
So that we don't have two conficting GovUK JS scripts loading, the Hale
theme stops the Block plugin from loading it's version when Hale and the
plugin run together.


MalcolmVonMoJ (approved):
Looks good to me. 👍

PR #61, opened 2021-04-22
Search screen title shading

MalcolmVonMoJ:
Added the blue shading for the search title and search bar.


RobLoweMOJ:
Just to ntoe the download buttons are amended in this PR. Will review on staging.

PR #89, opened 2021-08-02
Pluto styling tweaks

MalcolmVonMoJ:
- Added a gap betwixt the search box and the results heading.
- Added correct focus colouring for the breadcrumb links and other links with incorrect focus styling.

### DIFF
--- a/assets/scss/brandings.scss
+++ b/assets/scss/brandings.scss
@@ -257,16 +257,25 @@
 		#primary a:not([class]),
 		#primary a[class=""],
 		.hale-article-nav,
-		a.govuk-footer__link {
-
+		a.govuk-link,
+		a.govuk-footer__link,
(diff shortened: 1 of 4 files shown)


brown-a2 (approved):
Looks good. Question though, do we need to apply these changes to the other brandings as well?

RobLoweMOJ:
Copy of comments from slack: Since we now have this style in both neptune and pluto do you thinks its worth moving this to the brandings.scss I think keeping the breadcrumb focus state the same as the other branding focus states will be needed accross all brands/planets

MalcolmVonMoJ:
Agreed, these styles should be universal.  I've made the changes and removed the specific styling for Neptune from the specific brandings file.

brown-a2:
Looks good 👍

PR #123, opened 2021-12-21
WIP New colour approach

MalcolmVonMoJ:
- Split the long repetitive colour allocation into a new file
- Added new files for each planet, allocating a colour to each number
- moved Uranus, Neptune and Pluto to the new system.  
  - Venus and Earth left on the current system, albeit with the allocation in the new file.


MalcolmVonMoJ:
Colour work has been finished - this half-way approach was not needed

PR #134, opened 2022-01-06
Eris

MalcolmVonMoJ:
Enables planet Eris for use
- Adds Eris to the selectable list of planets
- Makes a few tweaks to the Eris brand to remove bugs and inconsistencies
  - Changes the coloured bar to Magistrates Turquoise - as that is who will use it.
  - Cookie banner colours.
  - Removes odd turquoise colouring to focussed elements for all brands, but which was only apparent in Eris.  
  - Introduces a few, not all, of the colour variables so existing styling tweaks work.
  - Extends `govuk-button` class for `wp-block-button__link`, for Eris only.

### DIFF
--- a/assets/scss/header.scss
+++ b/assets/scss/header.scss
@@ -196,9 +196,6 @@
     }
   }
 
-  .govuk-header__link:focus {
-    background-color: $colour_capri_teal;
-  }
   .govuk-header__navigation {
     border-top: 1px solid $colour_metal_grey_tint;
     li {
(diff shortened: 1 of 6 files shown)


brown-a2 (approved):

brown-a2:
One comment, Eris seems to hardly have any colour variables compared to the other brandings, is it supposed to?

MalcolmVonMoJ:
Eris is meant to use GDS colours - in theory it should have no colour variables whatsoever, but we have added a few as there are a few places where the GDS colours aren't being picked up.

PR #136, opened 2022-01-18
crown copyright logo added

MalcolmVonMoJ:
Added a style property to get the crown copyright logo.

### DIFF
--- a/footer.php
+++ b/footer.php
@@ -59,7 +59,7 @@ flush();
         if ('yes' == $crown_copyright) {
       ?>
         <div class="govuk-footer__meta-item">
-          <a style="background-image: url(<?php echo get_template_directory_uri(); ?>/assets/images/govuk-crest-2x.png)" class="govuk-footer__link govuk-footer__copyright-logo" href="https://www.nationalarchives.gov.uk/information-management/re-using-public-sector-information/uk-government-licensing-framework/crown-copyright/">© Crown copyright</a>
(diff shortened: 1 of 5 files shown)


footer.php
```diff
@@ -59,7 +59,7 @@
         if ('yes' == $crown_copyright) {
       ?>
         <div class="govuk-footer__meta-item">
-          <a class="govuk-footer__link govuk-footer__copyright-logo" href="https://www.nationalarchives.gov.uk/information-management/re-using-public-sector-information/uk-government-licensing-framework/crown-copyright/">© Crown copyright</a>
+          <a style="background-image: url(<?php echo get_template_directory_uri(); ?>/assets/images/govuk-crest-2x.png)" class="govuk-footer__link govuk-footer__copyright-logo" href="https://www.nationalarchives.gov.uk/information-management/re-using-public-sector-information/uk-government-licensing-framework/crown-copyright/">© Crown copyright</a>
```
RobLoweMOJ, line 62: Could we add this style to the govuk-footer__copyright-logo class rather then have it inline?

MalcolmVonMoJ: I'm not sure - because we get part of the url from a php function so I don't think we can do this in a stylesheet.

RobLoweMOJ: You can add to the sass file
`.govuk-footer__copyright-logo {
  background-image: url("../images/govuk-crest-2x.png")
}`

I did also have to add the following to the webpack mix file

`.copy('./assets/images/*', 'dist/images/')`

As the images should be in the dist folder

MalcolmVonMoJ: Done.  Thanks.

RobLoweMOJ (approved):

PR #139, opened 2022-01-20
Fix News Pagination so its accessible

RobLoweMOJ:


inc/pagination.php
```diff
@@ -24,19 +24,21 @@ function hale_archive_pagination($template = '')
...
+                    <li class="archive-pagination-current-page">
+                        <?php
+                        echo "Page " . $current_page_number . " of " . $max_pages; ?>
+
+                    </li>
+                    <li class="archive-pagination-prev-btn">
+                    <?php previous_posts_link('< Previous'); ?>
+                    </li>
+                    <li class="archive-pagination-next-btn">
+                        <?php next_posts_link('Next >'); ?>
```
MalcolmVonMoJ, line 38: I think the < & > signs should be done with CSS rather than a character.  But we can change this later.

MalcolmVonMoJ (approved):

PR #144, opened 2022-02-02
apply govuk font sizes

MalcolmVonMoJ:
Make the breakpoint for larger font sizes tablet so headings are consistent with body text.  
Use GDS font styling for body text and smaller headings - where size matches, using 19px for previous 18px.


RobLoweMOJ (commented):
Do we need to remove or adjust line heights? Does gov uk styles set these?

RobLoweMOJ (approved):

PR #146, opened 2022-02-07
Colour work

MalcolmVonMoJ:
- removed hard-coded colours in the CSS, replaced with GDS equivalents
- added more specific CSS selectors for the cookie banner in preparation for work on that plugin

### DIFF
--- a/assets/scss/brandings.scss
+++ b/assets/scss/brandings.scss
@@ -22,6 +22,11 @@
 		}
 	}
 }
+//remove gap at top of page
+.govuk-main-wrapper {
+	padding-top:0;
+}
+
 
 *{
 	@include branding($brandings) {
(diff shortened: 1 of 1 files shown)


assets/scss/news.scss
```diff
@@ -61,10 +61,10 @@
 
   .news-story-list-item-title {
     a {
-      color: $colour_metal_grey;
+      color: govuk-colour("dark-grey");
```
RobLoweMOJ, line 64: Are these the same colour? And if not is the branding settings setting it back to metal grey. Just making sure theres no visible change.

MalcolmVonMoJ: It is overridden, but this grey should probably be removed as:
- it is overridden by another style anyway
- it is a link, so grey is the wrong colour

So I'll kill it off.

RobLoweMOJ (approved):

PR #150, opened 2022-04-06
WIP Colour spike

MalcolmVonMoJ:


header.php
```diff
@@ -19,6 +19,7 @@
 	<meta name="theme" content="MOJ-hale-1.0.11">
 	<meta name="description" content="<?php bloginfo('description'); ?>" />
 	<link rel="profile" href="https://gmpg.org/xfn/11">
+  <link rel="stylesheet" id="custom-colour-css" href="https://jotwpublic.docker/wp-includes/css/MjB-test.css?time=<?php echo $_SERVER['REQUEST_TIME']; ?>" media="all">
```
MalcolmVonMoJ, line 22: Referencing the PHP-generated CSS file like this in the header.  Is this okay?

inc/customizer.php
```diff
@@ -363,6 +363,40 @@ function hale_customize_register( $wp_customize ) {
 		)
 	);
 
+	$wp_customize->add_setting(
+		'heading_colour',
+		array(
+			'default'           => '#0b0c0c',
+			'sanitize_callback' => 'sanitize_hex_color',
```
MalcolmVonMoJ, line 370: This sanitize callback will bean only hex values are allowed - I think this is a safer approach

inc/customizer.php
```diff
@@ -363,6 +363,40 @@ function hale_customize_register( $wp_customize ) {
...
+			'section'     => 'colors',
+			'type'        => 'text',
+		)
+	);
+
+	$wp_customize->add_setting(
+		'cookie_button_colour',
+		array(
+			'default'           => '#00703c',
+			'sanitize_callback' => 'hale_sanitize_nohtml',
```
MalcolmVonMoJ, line 387: This sanitize callback will mean colour names (e.g. "red") are also allowed, but if it doesn't match a colour then it will cause errors.  I think this is a more risky approach so I don't think we should do this one.

inc/template-functions.php
```diff
@@ -143,6 +143,24 @@ function hale_sidebar_location( $sidebar ) {
 
 function hale_get_branding_class() {
 
+	$css_file_path = "./wp/wp-includes/css/MjB-test.css";
```
MalcolmVonMoJ, line 146: This is the location for my test CSS file - it seems to work.  This is the Hale repository writing into the JOTW repository.  Is this okay?

inc/template-functions.php
```diff
@@ -143,6 +143,24 @@ function hale_sidebar_location( $sidebar ) {
...
+	$css_file_exists = file_exists($css_file_path);
+	$headingColour = get_theme_mod( 'heading_colour', '#0b0c0c' );
+	$cookieButtonColour = get_theme_mod( 'cookie_button_colour', '#0b0c0c' );
+
+	if ( $css_file_exists ) {
+		$css_file = fopen($css_file_path, "w") or die("Unable to open file!");
+		if (!empty($headingColour) ) {
+			$css = "h1, h2 {color:$headingColour!important;}";
+			fwrite($css_file, $css);
+		}
```
MalcolmVonMoJ, line 156: We can write the CSS file in sections like this, so any blank fields are ignored.

assets/scss/brandings.scss
```diff
@@ -55,6 +57,7 @@
 		#ccfw-page-banner {
 			.ccfw-banner__button:not(.ccfw-banner__button--close) {
 				background-color: branded(cookie-button-bg);
+				background-color: var(--cookie-button-background);
```
MalcolmVonMoJ, line 60: The non-IE11 method allows us to reference the new colour variable in the current CSS structure, keeping the new code in the current SCSS structure.

inc/customizer.php
RobLoweMOJ: I agree keeping it to hex is good approach

header.php
RobLoweMOJ: I will look at moving this to s3 bucket/uploads folder. But happy to discuss.

inc/template-functions.php
RobLoweMOJ: As mentioned I would look at moving this to s3 bucket/uploads folder. Mainly also because that would also mean theres only one true css file for each site to use. If the css files lives in the theme or includes folder each container running on AWS will have its own css file. Which could cause issues with some not being updated.

inc/template-functions.php
RobLoweMOJ: I wondering if we can do this in a string replace sort of way. So for example have css template with variables. That would mean rather then loads of if statements that may grow you just have an array of variables that get passed in. Each variable can have a default. Just thinking how we can make the code smaller. But happy to discuss.

inc/template-functions.php
RobLoweMOJ: I would also suggest we build the css file to a temp css file then once file is written maybe rename the css file to the right name for the site...just means if the css build does have an error the users wont see a half styled site.

assets/scss/specific-brandings.scss
```diff
@@ -1,103 +1,5 @@
...
-    Venus branding
-        VC
-**********************/
-.hale-branding--venus {
-	.govuk-header .govuk-header__logo {
-
-		a:focus img {
-			filter: invert(1) hue-rotate(180deg);
-		}
+.mojblocks-hero {
```
RobLoweMOJ, line 1: Can we move this so we can remove this sass file

functions.php
```diff
@@ -213,11 +213,44 @@ function hale_content_width()
...
  */
-function hale_scripts()
-{
+
+function action_customize_save_after( $array ) {
+    // Proof of concept,
+    // generate from options on page rather than preview CSS file to avoid editor clash of styles if someone else is previewing at the same time.
+
+    $upload_file_path = wp_get_upload_dir()["basedir"];
+    rename ($upload_file_path."/temp-colours.css", $upload_file_path."/custom-colours.css") or die ("Unable to rename file!");
```
RobLoweMOJ, line 222: Could add a if file_exists around each rename.

assets/scss/specific-brandings.scss
MalcolmVonMoJ: Done - moved to moj-blocks-general.scss

inc/template-functions.php
```diff
@@ -141,26 +141,15 @@ function hale_sidebar_location( $sidebar ) {
 	return $sidefloat;
 }
 
+require get_template_directory() . '/inc/colour-branding.php';
```
RobLoweMOJ, line 144: Move to functions or inc/colours.php

inc/colour-branding-import.php
```diff
@@ -0,0 +1,22 @@
...
+	$jason_file = get_theme_mod("customizer_setting_json");
+	if ($jason_file) {
+		$jason_string = file_get_contents($jason_file);
+		$jason = json_decode($jason_string, true);
+		$colour_array = hale_get_colours() or die("no colour array");
+
+		for($i=0;$i<count($colour_array);$i++) {
+			$colour_id = hale_get_colour_id($colour_array[$i]);
+			$jason_colour = $jason[$colour_id];
+			if (get_theme_mod($colour_id) != $jason_colour) //if JSON colour is different from current colour
```
RobLoweMOJ, line 13: It it worth doing a validation here to make sure its a valid hex colour?

functions.php
MalcolmVonMoJ: Done!

inc/colour-branding.php
```diff
@@ -0,0 +1,121 @@
...
+/************** 
+ * This file creates two CSS files
+ * One with the variables that are set in the Customiser, this also has the colour bar style and the invert logo focus style
+ * One for IE which copies the entire main CSS file and replaces the variables with the colours
+ * SVG support however means that the IE CSS file is needed by every browser at the moment - this might change in the future
+ * IE doesn't support the filter attribute, so the logo focus filter is only appended to the non-IE CSS file
+**************/
+
+// JSON import code
+require get_template_directory() . '/inc/colour-branding-import.php';
```
RobLoweMOJ, line 12: The export is in customizer.php. We should keep colour files imports together

inc/colour-branding-import.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+function hale_get_colours_from_jason() {
+	$jason_file = get_theme_mod("customizer_setting_json");
```
RobLoweMOJ, line 4: Maybe worth being abit more descriptive with variable names rather then just $json/jason .. e.g. $custom_colours_json

inc/colour-branding-import.php
MalcolmVonMoJ: No.  
- Wordpress does this check for us.  
- One of the values is not a hex value, but can be raw CSS.  We should remove this eventually but for now it has to be there.

inc/template-functions.php
```diff
@@ -141,27 +141,7 @@ function hale_sidebar_location( $sidebar ) {
...
-
-    if ( !empty($colour) ) {
-        $theme_colour_name = 'hale-branding--' . $colour_array[ $colour ];
-        return $theme_colour_name;
-    } else {
-        return 'hale-branding--neptune'; //none set = use Neptune
-    }
-
-}
+require get_template_directory() . '/inc/colour-branding.php';
```
RobLoweMOJ, line 144: Move to colours.php or functions.php

brown-a2:
Happy for this to get pushed to the pipeline once all the comments here are resolved.

brown-a2 (approved):

RobLoweMOJ (approved):

PR #151, opened 2022-04-08
Add Submenu functionality and styles

RobLoweMOJ:
I have added some branding styles but will expand this to other link states mobile/desktop in next PR. We may want to think about how we standardise the name of our variables.

### DIFF
--- a/assets/js/navigation.js
+++ b/assets/js/navigation.js
@@ -150,3 +150,21 @@ function guideNavClick(id) {
 		document.getElementById(id).style.display = "block";
 	}
 }
+
+jQuery( document ).ready(function( $ ) {
+  $( "#menu-menu-top-menu li.menu-item-has-children > a" ).click(function( event ) {
+
+    if($('.govuk-header__menu-button').is(':visible')) {
+
+      event.preventDefault();
+
+      if ($(this).parent().hasClass('sub-menu-open')) {
+        $(this).parent().removeClass("sub-menu-open");
+      } else {
+        $("#menu-menu-top-menu li.menu-item-has-children").removeClass("sub-menu-open");
+        $(this).parent().addClass("sub-menu-open");
+      }
+    }
+
+  });
+});
--- a/assets/scss/header.scss
+++ b/assets/scss/header.scss
@@ -38,9 +38,7 @@
 
           &:focus {
             &:before {
-              border: solid govuk-colour("black");
-              border-width: 0 3px 3px 0;
-              padding: 3px;
+              border-color: govuk-colour("black");
             }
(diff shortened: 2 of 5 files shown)


assets/js/navigation.js
```diff
@@ -150,3 +150,21 @@ function guideNavClick(id) {
...
+    if($('.govuk-header__menu-button').is(':visible')) {
+
+      event.preventDefault();
+
+      if ($(this).parent().hasClass('sub-menu-open')) {
+        $(this).parent().removeClass("sub-menu-open");
+      } else {
+        $("#menu-menu-top-menu li.menu-item-has-children").removeClass("sub-menu-open");
+        $(this).parent().addClass("sub-menu-open");
+      }
```
MalcolmVonMoJ, line 166: Would `.toggleClass()` work here?

assets/scss/header.scss
```diff
@@ -4,18 +4,162 @@
...
+              border-width: 0 3px 3px 0;
+              padding: 3px;
+            }
+          }
+        }
+
+        &.sub-menu-open {
+          > a {
+            &:before {
+              transform: rotate(-135deg);
```
MalcolmVonMoJ, line 51: We need a `-webkit-transform` here too, to match line 30.

MalcolmVonMoJ: Alternatively, you could change the border-width to be opposite from line 26.

assets/scss/header.scss
```diff
@@ -4,18 +4,162 @@
...
+            @media (min-width: 48.0625em) {
+              content: none;
+            }
+          }
+
+          &:focus {
+            &:before {
+              border: solid govuk-colour("black");
+              border-width: 0 3px 3px 0;
+              padding: 3px;
```
MalcolmVonMoJ, line 43: Are these needed?  They are the same as lines 26 and 28.

MalcolmVonMoJ (changes requested):
Looks good, the only think I think _needs_ to be added is the `-webkit-transform` for the menu open state.

assets/js/navigation.js
RobLoweMOJ: I originally used this but because you need to do the removeClass on all li elements when only when opening a submenu I thought an if else would be more suitable. Unless you can see a way round it.

assets/scss/header.scss
RobLoweMOJ: Go suggestion. i think i change line 41 to border-color then i wont need the other lines.

MalcolmVonMoJ (approved):
Looks good

PR #154, opened 2022-05-05
Disable archive pages (except news)

brown-a2:
Archive pages load by default even if they are not used. We don't want
these pages loading and being picked up by search engines,
until we have a use for them and have them styled.


MalcolmVonMoJ (approved):
Looks good to me

PR #170, opened 2022-06-17
unset headings colour

MalcolmVonMoJ:
Unset h1-h6 colour for accordion so headings will be black.  No actual change for h4-h6 but included for completion.


MalcolmVonMoJ:
It reverts to default body text - inheriting from `.govuk-body` which it is within.

RobLoweMOJ:
Approved but we need to look at heading colours in general as they are not gds black

RobLoweMOJ (approved):

PR #171, opened 2022-06-20
Colour preset

MalcolmVonMoJ:
New code to account for new colour settings being added without the colour being specified:
- every time the page loads it does a time check betwixt `custom-colours.css` and `colours.php`
  - if the css file is newer than the php file, then the custom colours code would have picked up all new colours and all is well
  - otherwise:
    - we do a check for any new colours and add them to the end of the css file - set to their default
    - if there are no extra colours, we still touch the css file to mark it as checked, so we don't have to search for new colours again.
- This still assumes that new colours are added in the correct way, and included in `colours.php`

### DIFF
--- a/inc/colours.php
+++ b/inc/colours.php
@@ -174,4 +174,38 @@
 	function hale_get_colour_options($colour_array_item) {
 		return $colour_array_item[4];
 	}
-?>
+
+	function hale_new_colour_check() {
+		// This function checks for any newly created colours that haven't been set in the customizer and applies the default.
+		// The amended CSS file will be overwritten when the colours are next amended.
+		// This does not protect against new colours being added to the SASS and not added to colours.php.
+		// This is for non-IE only as this particular bug doesn't occur in IE.
+
+		$CSSfileURL = wp_get_upload_dir()["basedir"]."/custom-colours.css"; //non-IE CSS file (variable values)
+		$ColourFileURL = get_template_directory()."/inc/colours.php"; //colours CSS file
+
+		if (filemtime($CSSfileURL) <= filemtime($ColourFileURL)) {
+			//The colours CSS for this site is older than the last colours build - doesn't necessarily mean that the CSS is incomplete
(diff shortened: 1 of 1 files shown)


inc/colours.php
```diff
@@ -174,4 +174,42 @@ function hale_get_colour_hint($colour_array_item) {
...
-?>
+
+	function hale_new_colour_check() {
+		// This function checks for any newly created colours that haven't been set in the customizer and applies the default.
+		// The amended CSS file will be overwritten when the colours are next amended.
+		// This does not protect against new colours being added to the SASS and not added to colours.php.
+		// This is for non-IE only as this particular bug doesn't occur in IE.
+
+		$CSSfileURL = wp_get_upload_dir()["basedir"]."/custom-colours.css"; //non-IE CSS file (variable values)
+		$ColourFileURL = get_template_directory()."/inc/colours.php"; //colours CSS file
```
RobLoweMOJ, line 185: Should this be a css file not the php file?

MalcolmVonMoJ: The PHP file contains the latest colours, so we need to check the CSS file (184) with the PHP file (185).  If the CSS file is newer, then the code that created it ran with the latest colours so we can be confident that it is up-to-date.

RobLoweMOJ (commented):
Does the ie css file also need to be regenerated or will we raise a sep ticket for this?

RobLoweMOJ:
Looks good. As discussed on friday I think an improvement would be the the trigger for check would be by version number change (preferably the theme vers number) rather than the file time. I would need to check but when aws spins up new containers to it could be the files are newer so this could mean this could run when its not needed to. This colour check really only needs to only run after we deploy on the pipeline.

RobLoweMOJ (approved):
Great work. Happy to approve. Though we can look at improvements.

MalcolmVonMoJ:
Yes, let's kick the version number stuff into a different ticket.  
I'll go ahead and merge this then.

PR #177, opened 2022-07-15
GDS Design System update to 4.2.0

MalcolmVonMoJ:
- Imported the GDS Design System stuff and corrected some breaking changes
  - Removed our current `govuk-frontend.js` file and replaced with a copy from `node_modules` in our `webpack.mix.js`
  - New accordion style changes (Accordion block in moj-blocks, colour changes are here)
    - section-heading no longer styled as a link
    - expanded colours to the new chevron for the show/hide link
    - removed previous little triangle
    - fixed accordion bugs
      - Duplicate IDs existed in the accordion, these are now changed in JavaScript
      - When a link direct-links to an ID on the page, the accordion will now open that section
        - Note: this done in raw JavaScript, not jQuery, as it has been shared with GDS with a (possible) aim of improving the Design System code - which does not use jQuery
  - Header
    - Added `display: inline-block;` style to logo (class `govuk-header__link--homepage`) so focus state is still applied and visible
      - GDS were fixing an issue with theirs that we had already fixed in a different way
    - Navigation class rename, the old `govuk-header__navigation` is now `govuk-header__navigation-list`
    - The `<nav>` element now has the old class `govuk-header__navigation`
    - The `<button>` for the menu now needs to appear inside the `<nav>` rather than before, so it has been moved from `header.php` to `partials/topnav.php`
  - The skip link at the top has a new data module
- Secondary nav changes
  - Changed the padding to only be for widths of 1024px or greater to prevent unwanted wrapping (mainly in CCRC) at narrower breadths
  - Changed some of the colour defaults to prevent any unwanted change (Magirec) to current page style


brown-a2:
Have you tested on a page where multiple accordions have been added? Just to check the looping works as expected.

MalcolmVonMoJ:
Yes, I've tested on this page:
https://ccrc.gov.uk/who-we-are/

(Well, my local version of it)

brown-a2:
👍

brown-a2 (approved):

RobLoweMOJ (commented):
Looks good. Though have a couple of questions around the auto expand. Lets catchup about it tomorrow.

PR #189, opened 2022-10-21
Document Listing Page

RobLoweMOJ:


brown-a2 (approved):
Looks good. Nothing breaking that I can see. My only issue is that there is a lot of repetition going on in the page-document-listing.php page, with the same code being repeated 3 times for each doc type, multiple times.

PR #213, opened 2023-05-19
News listing and article - margins

MalcolmVonMoJ:
Added margins to published date
Added bottom margin to featured image
Re-specified top margin for featured image in rem
Changed category tag margins to work better with the new margins for published date

### DIFF
--- a/assets/scss/news.scss
+++ b/assets/scss/news.scss
@@ -13,9 +13,13 @@
 }
 
 .news-story-featured-image {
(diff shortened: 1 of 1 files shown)


brown-a2 (approved):

brown-a2:
Does it matter whether tags are being used or not for this?

MalcolmVonMoJ:
It shouldn't, but I will do a few extra tests to be sure.

PR #222, opened 2023-07-07
Accessibility improvements

MalcolmVonMoJ:
Fixes various accessibility failings/observations as per Martin Glancy's comments:

### R14
Double H1 on search results page when none found
- Changed the second H1 to an H2.  
- Changed the class to a suitable class.
- Amended CSS as GDS heading class styling was being overridden.
- Added a few more classes elsewhere to prevent style changes by above change. 

### R13
Search box covers up banner text (page title/logo), also the search box breaks apart at some widths.
- Search box switches to mobile earlier (same as menu)
- Max-width styles introduced to stop search box from breaking up - this means that the site title might wrap a little earlier now.

### R06
Horizontal scrolling caused by earlier fix to highlighted fields
- Added a style to prevent horizontal scrolling, as there is no content outside of the visible area this is all that is needed.  

### R15
Search for term not clear
- Design decision, but I have increased contrast here to improve matters. 
- I did this by changing it from a `caption-m` to a `body-l`. 
- Rejig of HTML to get this to display as needed, especially as we want both lines to be within the H1.

### R16
Ghost links in search results
- This was caused by HTML in page titles, specifically links.  Whilst this is a content problem, to increase system robustness, I have stripped all HTML tags from the search result title.  

### R24
Blank searches return results
- Not a problem per se, but the search page just says "search results for", which isn't great. 
- I have added logic so if the search term is blank, it tells the user that no search term was entered.  
- Wording approved by Anna.

### Also
- Future proofed media queries in footer by making them use the GDS breakpoints.

### DIFF
--- a/assets/scss/header.scss
+++ b/assets/scss/header.scss
@@ -272,14 +272,17 @@
     }
   }
 
-  .govuk-header__logo {
-    //make this a min-width to account for the longer header texts than "GOV.UK"
-    width: unset;
-    min-width: 33.33%;
+  &.hale-header--with-search .govuk-header__logo {
     @include govuk-media-query($from: desktop) {
       // space for search box
       max-width: calc(100% - 285px);
     }
+  }
+
+  .govuk-header__logo {
+    //make this a min-width to account for the longer header texts than "GOV.UK"
+    width: unset;
+    min-width: 33.33%;
 
     .govuk-header__link--homepage {
       display: inline-block;
--- a/assets/scss/header.scss
+++ b/assets/scss/header.scss
@@ -276,6 +276,10 @@
     //make this a min-width to account for the longer header texts than "GOV.UK"
     width: unset;
     min-width: 33.33%;
+    @include govuk-media-query($from: desktop) {
+      // space for search box
+      max-width: calc(100% - 285px);
+    }
 
     .govuk-header__link--homepage {
       display: inline-block;
@@ -296,6 +300,7 @@
 
       .hale-header__logo--custom {
         float: left;
+        min-width: 43px;
       }
     }
   }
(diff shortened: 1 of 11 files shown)


assets/scss/header.scss
```diff
@@ -276,6 +276,10 @@
     //make this a min-width to account for the longer header texts than "GOV.UK"
     width: unset;
     min-width: 33.33%;
+    @include govuk-media-query($from: desktop) {
+      // space for search box
+      max-width: calc(100% - 285px);
+    }
```
brown-a2, line 283: Does this account for when someone doesn't add a search box in the header?

assets/scss/moj-blocks-general.scss
```diff
@@ -1,3 +1,11 @@
+.hale-page {
+  // Deals with the Hero and the highlight/shaded blocks (e.g. CTA, highlighted list)
+  // The shading for these have (need) widths of over 100vw which leads to undesirable horizontal scrolling
+  // No content is outside of the page area, just highlighting.
+  // This prevents this from being an issue by preventing horizontal scroll
+  overflow-x: hidden;
+}
+
```
brown-a2, line 8: Can you reuse one of the default WP generated classes like `hale-header` or if it's specifically for a page or site, `page-header-default` /  `hale-site-5`?

brown-a2, line 8: Is this just something that will effect those two blocks or do you think it is worth applying to the wider site?

brown-a2 (approved):

assets/scss/header.scss
MalcolmVonMoJ: No, the header will be restricted as if there is a search box.  I'll have a think about this.  Thanks.

assets/scss/moj-blocks-general.scss
MalcolmVonMoJ: This will apply to the wider site.  If there are other blocks that have content (as opposed to just styling) disappearing off the edge, that block will either need to be able to horizontal scroll itself or another fix will be needed.  I couldn't find any such issue at the moment.

assets/scss/header.scss
MalcolmVonMoJ: I have changed the CSS so this restriction is only when search boxes are shewn.

PR #224, opened 2023-07-18
Menu orientation [WIP]

MalcolmVonMoJ:
**WIP**
There are still a few things to do on this at the moment.

Changed the CSS to orientate the second-level navigation to be vertical instead of horizontal.  

- Switch display flex to display block.
- Positioning changes to ensure the menu starts at the right place (just underneath the parent item).
- Set the new default colours for the new secondary nav (replacing the defaults for the old one).
  - All sites (except Magistrates which uses the defaults) will need to re-examine their subnav colours to ensure they are still okay, this is most pressing for IMB, which used a mix of their own colours and defaults.  Not for this PR.
- Removed the spacers that were used to ensure there was space on the screen for the old nav (no longer needed as nav only pops up when the parent is hovered over).
- Left aligned menu items so they're in line with the new subnav.
- Updated the underline mixin to enable it to be applied at different widths.
  - Primary nav - from 1024
  - Secondary nav - from GDS desktop (769)
- Rejigged some of the mobile colours as needed, to ensure consistency and correct problems introduced by the new CSS
  - Pulled some styling out of desktop only into general
  - Moved a few colours around.
- Added `margin-right:-40px` to subnavs of the last primary menu item, to deal with edge-cases where words might be broken asunder.
- Fixed a few outstanding issues
  - Spelling error in colours.php
  - Inconsistent bottom border if last menu item had children (mobile only)
  - Ensured keyboard tabbing could reach the subnav (I can't believe no-one noticed that this didn't work!)


MalcolmVonMoJ:
This has been incorporated into PR #227

PR #225, opened 2023-07-26
Footer languages

MalcolmVonMoJ:
- Added a new menu option for menus for language attributes.
  - But hid it unless revealed (as vast majority of links will not be in a different language).
  - Added examples of the ISO codes.
- Added the lang attribute to main and secondary footer menus.

Also added `<span>` tags to the top menu.  This makes no difference now, but it is used in other menus on Hale already and might make future styling easier.


RobLoweMOJ (approved):
Looks good. Though i would recommend adding hale to the function names and maybe creating a new file in inc for this so the functions file doesnt get too big.

PR #228, opened 2023-08-04
Updated FontAwesome to use new Twitter logo

MalcolmVonMoJ:
FontAwesone have released a new version with the new Twitter logo.  

I have:
- Downloaded the new font files
  - There are no longer any `eot`, `woff` or `svg` font files, only `ttf` and `woff2`.  I have deleted the obsolete ones.  
- Downloaded the new SCSS files
  - The SCSS file `_larger.scss` is no longer part of the package, so I've deleted this.
  - New SCSS files: `_sizing.scss` & `_functions.scss`
- Changed `icons.scss` to use the new X twitter logo

In addition to this, I have added an option for sites to use the non-harmonized (non-square) logos.  A new option in `social-widget.php` switches betwixt the square versions and the non-square versions.


RobLoweMOJ (commented):
As mentioned please could you check this change on browserstack  as the removal of eot, woff or svg may effect things. Also if the default is non square we need to account for that when we deploy and update the sites.

MalcolmVonMoJ:
IE11 is okay.  
Yes, the roll-out will need the harmonized box to be ticked for the sites which currently use this.  
- CCRC
- Magistrates
- VC

RobLoweMOJ (approved):
LGTM

PR #229, opened 2023-08-11
Card Block background colour

MalcolmVonMoJ:
The card background has been hard-coded to white.  This adds in an option to change this to match designs as colours shouldn't be hard coded.  No colour-refresh should be needed.


MalcolmVonMoJ:
This change doesn't remove the hard-coding as that is in [the mojblocks repo](https://github.com/ministryofjustice/wp-moj-blocks/blob/9915f946261b8ec81b5f426f6c467c401c05f5a1/src/custom-blocks/card/style.scss#L14).  However, the CSS of this change will take priority and the hard-coded colour can be removed later.

brown-a2 (approved):

PR #233, opened 2023-08-17
Add ACF field to restrict job listing page by organisation

RobLoweMOJ:


MalcolmVonMoJ (changes requested):
This is quite old now.  I think a lot of these changes have since been done elsewhere - sometimes the same, sometimes different as per changing requests.  For example, the page size was upped to 50 rather than 25.  

If this is still needed, I suggest the changes in `assets/scss/job-listings.scss` and `page-job-listing.php` are reverted here before merging - so essentially only `acf-json` changes are pushed.

brown-a2:
Agreed, we'll get @RobLoweMOJ to review

PR #234, opened 2023-08-25
Small changes to job listing page

MalcolmVonMoJ:
Page size set to 50 and added comma to LG for capitalisation.

### DIFF
--- a/inc/job-listing.php
+++ b/inc/job-listing.php
@@ -9,7 +9,7 @@
 	 **/
 
 	function sanitizeAddress($address) {
-		$allCaps = ["HM","HMP","YOI","HMP/YOI","HMYOI","HMPYOI","NPS","NMS","RCJ","OPG","JCP","LAA","HMCTS","HMPPS","CTS","LG"];
+		$allCaps = ["HM","HMP","YOI","HMP/YOI","HMYOI","HMPYOI","NPS","NMS","RCJ","OPG","JCP","LAA","HMCTS","HMPPS","CTS","LG,"];
 		$specialCases = ["&amp;", "MoJ", "(MoJ)", "MacMillan"]; //instances where the capitalization isn't quite normal
 		$commonSeparators = [
 			// These are separators which are not capitalised at all
--- a/page-job-listing.php
+++ b/page-job-listing.php
@@ -10,7 +10,7 @@
 get_header();
 
 //Get Search Filter Values
-$page_size = 12;
+$page_size = 50;
 $selected_job_role_id = 0;
 $selected_job_region_id = 0;
 $selected_job_min_salary_id = 0;


inc/job-listing.php
```diff
@@ -9,7 +9,7 @@
 	 **/
 
 	function sanitizeAddress($address) {
-		$allCaps = ["HM","HMP","YOI","HMP/YOI","HMYOI","HMPYOI","NPS","NMS","RCJ","OPG","JCP","LAA","HMCTS","HMPPS","CTS","LG"];
+		$allCaps = ["HM","HMP","YOI","HMP/YOI","HMYOI","HMPYOI","NPS","NMS","RCJ","OPG","JCP","LAA","HMCTS","HMPPS","CTS","LG,"];
```
brown-a2, line 12: Did you mean to add the comment in the end here?

brown-a2 (approved):

inc/job-listing.php
brown-a2: Technically, shouldn't cause an issue, but just wondering as the other array in the file doesn't have it.

brown-a2:
Also... this is completely a style thing, so feel free to ignore. I might be tempted to swap out the if conditionals with a switch statement. It feels a bit cleaner but then again maybe not 🤔 Anyway, it would look something like this...
```
foreach ($addressBits as $bit) {
    switch (true) {
        case in_array(strtolower($bit), array_map('strtolower', $specialCases)):
            // This ensures that special cases are capitalized correctly
            foreach ($specialCases as $specialCase) {
                if (strtolower($bit) == strtolower($specialCase)) {
                    $sanitizedAddress .= $specialCase . " ";
                }
            }
            break;
        
        case preg_match("/[0-9][SNRTsnrt][TDHtdh]/", $bit):
            // This ensures items like 1st/2nd are lower case
            $sanitizedAddress .= strtolower($bit) . " ";
            break;

        case in_array($bit, $allCaps):
            // This ensures all-caps items are capitalized
            $sanitizedAddress .= strtoupper($bit) . " ";
            break;

        case !preg_match("/[0-9]/", $bit):
            // This is everything else if no numbers found in string
            $bit = mb_strtolower($bit);
            if (!in_array($bit, $commonSeparators)) {
                $bit = mb_convert_case($bit, MB_CASE_TITLE) . " ";

                // The below deals with capitalization in places like "Wells-next-the-Sea" (hyphens)
                foreach ($commonSeparators as $separator) {
                    // First, capitalize the "Sea" bit, which would have been missed before as it is not at the beginning of the "word"
                    $disectedTownName = explode("-$separator-", $bit);
                    if (count($disectedTownName) == 2) {
                        $disectedTownName[1] = mb_convert_case($disectedTownName[1], MB_CASE_TITLE);
                        $bit = join("-$separator-", $disectedTownName);
                    }
                }
            }
            $sanitizedAddress .= "$bit "; // Construct the sanitized address

            foreach ($commonSeparators as $separator) {
                // Here, we decapitalize any separators
                $sanitizedAddress = str_ireplace("-$separator-", "-$separator-", $sanitizedAddress); // Hyphens
                $separator = str_replace("-", " ", $separator);
                $sanitizedAddress = str_ireplace(" $separator ", " $separator ", $sanitizedAddress); // Spaces
            }
            break;

        default:
            $sanitizedAddress .= $bit . " ";
            break;
    }
}```

inc/job-listing.php
MalcolmVonMoJ: As it is in the quotes it just is part of the string, so should be okay.  It seems to work as intended.

MalcolmVonMoJ:
A switch statement might be cleaner.  But I don't think I should do it here as it will dwarf the other changes in this PR.

PR #240, opened 2023-09-06
Change heading size

MalcolmVonMoJ:


brown-a2 (commented):
Is there a missing comma after $id ? Should it be? 
```
<div class="job-list-item" id="job-<?php echo $id; ?>">
    <h2 class="job-list-item--title govuk-heading-m">
        <a class="govuk-link" href="<?php printf($url); ?>">
            <?php echo get_the_title(); ?>
        </a>
    </h2>
</div>
```

PR #243, opened 2023-09-07
New Nav Bar Styles

MalcolmVonMoJ:
New navigation bar styles:
- Nav bar items are displayed in groups of 4, 5, 6 & (1024px or wider) 7.
  - New data attribute `data-menu-count` which is added to the HTML by PHP.
  - This is handled by their width attributes (`calc(0.88 * (100% - 50px)` divided by the number per row)
  - A gap is kept to the right to allow dropdown menus to fill the space. 
- Removed the `space-between` flex style - now all items are bunched together to the left and wrap as needed.    
- They are left aligned.
- The padding around them is no longer removed when screen is narrower than 1024px.
- Submenu items (drop downs) are at least the width of their parent and can grow slightly.
- Submenu items have a little less padding to make them in line with their parent.
  - Mixin has a new variable `$gap` to handle this.
- Main nav bar is 28px wider and shifted 14px left.  This puts the menu text in line with the rest of the content.
- SCSS Desktop styles are all grouped together for easier management.


brown-a2 (approved):

brown-a2:
Codewise looks fine. In terms of the results, really not convinced but not up to me.

PR #246, opened 2023-09-18
Revert "New Nav Bar Styles"

MalcolmVonMoJ:
Reverts ministryofjustice/wp-hale#243

### DIFF
--- a/assets/scss/brandings.scss
+++ b/assets/scss/brandings.scss
@@ -1,14 +1,14 @@
 @import 'page-colours';
 
 //mixins
-@mixin hale-heading-link-underline-restrictor($colour, $width: desktop, $gap: 15px) {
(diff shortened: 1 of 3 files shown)


assets/scss/brandings.scss
```diff
@@ -320,7 +329,7 @@
 								> a:not(:focus):link:visited {
 									background-color: var(--header-link-with-children-hover-bg);
 									color: var(--header-link-with-children-hover);
-									border-color: var(--header-link-with-children-hover-border);
+									border-color: var(--header-link-with-children-hover-border);;
```
RobLoweMOJ, line 332: We could keep this removed

RobLoweMOJ (approved):

assets/scss/brandings.scss
MalcolmVonMoJ: Done

PR #247, opened 2023-09-20
New nav menu [WiP - pending UR feedback]

MalcolmVonMoJ:
New navigation bar styles:
- Nav bar items are displayed in groups of 4, 5, 6 & (1024px or wider) 7.
  - New data attribute `data-menu-count` which is added to the HTML by PHP.
  - This is handled by their width attributes (`calc(0.88 * (100% - 50px)` divided by the number per row)
  - A gap is kept to the right to allow dropdown menus to fill the space. 
- Removed the `space-between` flex style - now all items are bunched together to the left and wrap as needed.    
- They are left aligned.
- The padding around them is no longer removed when screen is narrower than 1024px.
- Submenu items (drop downs) are at least the width of their parent and can grow slightly.
- Submenu items have a little less padding to make them in line with their parent.
  - Mixin has a new variable `$gap` to handle this.
- Main nav bar is 28px wider and shifted 14px left.  This puts the menu text in line with the rest of the content.
- SCSS Desktop styles are all grouped together for easier management.

Also the below, which were not part of the since-reverted PR #243, but iterated later. 
- Removed dropdown arrows
- Added new colour options:
  - Underline for all menu items
  - Background for all menu items
  - Bottom border for main menu
  - Remove dropdown arrows
- This will need new colours to be set for every site - a simple colour refresh won't cut it.


brown-a2 (approved):

MalcolmVonMoJ:
This is no longer needed as the later PR (#277) has solved these issues in a more flexible way.

PR #254, opened 2023-09-22
Job data attribute

MalcolmVonMoJ:
This adds the missing `data-location-id` attribute to each job with a prison in the prison list


brown-a2 (approved):

brown-a2:
You know how I feel about loops inside loops! But I guess I'll look the other way this time. Aside from that, the only other thing, do we have to use that fixed list of prisons, can't we just extract the name from the location/address so we don't have a list to maintain in here?

MalcolmVonMoJ:
I'm not sure how we can automate it as the list seems to be too diverse.  For example, this entry - I'm not sure how we will be able to automate this as being one entry where other multi-prison jobs are two.  

Also there is which name to use - some prisons are "officially" HMP, others are "HMP/YOI" - I don't know how we'd automate that without a master list like this.  
```
			{
			   "name":"HMP Grendon\/Springhill",
			   "town":"Aylesbury",
			   "lat":"51.8931365",
			   "lng":"-1.0072932",
			   "type":[
				  "Prison"
			   ],
			   "name_variations":[
				  "HM Prison Grendon",
				  "HMP Grendon",
				  "HM Prison Springhill",
				  "HMP Springhill"
			   ]
			},
```

PR #263, opened 2023-10-20
Footer white bar fix

MalcolmVonMoJ:
This is a quick fix to the footer bug where a white bar is shewn at the bottom of the screen.  

Bug is caused by a seemingly invisible image added after the footer - bottom of body.

Fixed in 2 ways:
- `box-shadow` to extend beyond the footer to cover up any bar
- `display:none` to hide the 1×1px image which is added by Facebook cookies that causes this issue

The first of these will prevent these sorts of issues, assuming the code adds them at the bottom of the page after the footer.  
The second will work if it is an image added just beneath the body at any point.  
Both will work in this situation.

### DIFF
--- a/assets/scss/brandings.scss
+++ b/assets/scss/brandings.scss
@@ -681,6 +681,7 @@ html.hale-colours-variable {
 		border-top-width: 4px;
 		border-top-color: var(--footer-border);
 		background: var(--footer-background);
+		box-shadow: 0 20px 0 0 var(--footer-background);
 
 		a:not([class]),
 		a[class=""],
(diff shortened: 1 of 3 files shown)


RobLoweMOJ (approved):
Worth checking if FB tracking still works after deployment

MalcolmVonMoJ:
Looking at other similar pixels on our other sites, they have used `display:none` to hide themselves, therefore I am pretty sure that this approach is okay.  However, I have talked with Marc and we will be vigilant about any drop-off. Will wait until tomorrow before deploying.

RobLoweMOJ (approved):

PR #268, opened 2023-11-02
removed .mo files from git ignore list

MalcolmVonMoJ:


MalcolmVonMoJ:
Not needed, will do elsewhere

PR #274, opened 2023-11-29
New tab link suffix

MalcolmVonMoJ:
- New CSS style to label all links with `target="_blank"` as opening in a new tab.
  - The new style uses a PHP variable, so this is in the `<head>` tag rather than in an SCSS file.  
- New CSS style for edit page view, so editors can see this will have a suffix.  
  - This is a static bit of text "(new tab link)" so this can be in an SCSS file.  
  - This has an opacity of 0.6 so editors can see it is not an editable bit of text.
- Adds setting to customise the suffix, enabling different sites to use different text if needed.
  - It was noted that PPJ would use "open in a new tab" and NPM would use "open in a new window".
  - This is free text so can be anything.
  - This is a new section "Links".

## For rollout:
- The new setting won't be set yet on any page, so no links will have the suffix.  
- Someone will need to go through each page on a removing the manual "opens in a new tab" suffix.  
- Only once they are all removed can the text can be set in the new setting, otherwise the new automatic suffix will appear alongside the old suffix.

### DIFF
--- a/assets/scss/editor.scss
+++ b/assets/scss/editor.scss
@@ -36,11 +36,4 @@
 			border-bottom-style: none;
 		}
 	}
-
-	// Link suffixes
-	a[target=_blank]:after {
-		content:" (new tab link)";
-		opacity: 0.6;
-	}
-
 }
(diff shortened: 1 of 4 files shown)


RobLoweMOJ (approved):
Approved but as discussed we should move to settings as part of moving to site editor

PR #279, opened 2023-12-19
Fix page-document-listing template missing var bug

brown-a2:
When searching documents, if you left the search field blank you would get a missing var warning. I've added an empty string to fix this so var always set to something.


MalcolmVonMoJ:
Do we need the same protection in `page-job-listing.php`?

MalcolmVonMoJ (approved):

PR #280, opened 2023-12-28
Search term retention

MalcolmVonMoJ:
Added a new `value` property to the search form and used the existing `$query` value to fill it.  Changed the no search term `$query` value from "Search" to "" so the value is blank should no search term be entered.  

It looks like `$query` was not used - perhaps a hang-on from a previous incarnation of the search page.  But I thought it made sense to use it here.

### DIFF
--- a/searchform.php
+++ b/searchform.php
@@ -12,7 +12,7 @@
 if ( ! empty( get_search_query() ) ) {
 	$query = get_search_query();
 } else {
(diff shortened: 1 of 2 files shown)


brown-a2 (dismissed):
But you'll need to increase the version one more as my last merge used the version number you're using. Just override and merge once you change that.

PR #281, opened 2024-01-03
Add CCRC Decision custom post type to theme

brown-a2:
On the request of CCRC, we've added a new custom post type to help them manage decisions and record them on the site. They have the ability to create a decision post and build a landing page or deicison archive to display and filter all the decision posts based on several taxonomies.

Changes/addisions are:

* Adds a new custom post type called "decision" and section in lefthand menu called "Decision Archive".
* Adds ability to toggle on/off in customizer.
* Adds 3 new taxonomies - Referral Type, Offence, Process
* Adds a new decision single page.
* Adds a new Decision template to support new post type. Creates a landing page where posts can be filtered.
* Adds new ACF fields used by the new template.
* Added new CSS to support the new custom post type template page.


assets/scss/decisions.scss
```diff
@@ -0,0 +1,115 @@
...
+            height: 1em;
+            width: auto;
+        }
+    }
+
+    .decision-summary-text p {
+        font-weight: bold;
+        margin-bottom: 20px;
+    }
+}
```
MalcolmVonMoJ, line 62: I don't think any of this is needed.  I can't find where it would be applied.

page-decision-listing.php
```diff
@@ -0,0 +1,347 @@
...
+                }
+
+                $decision_query = new WP_Query($decision_args);
+                $decision_referral_type_filter_activated = get_post_meta(get_the_ID(), 'decision_referral_type_filter_activated', true);
+
+                if ($decision_query->have_posts()) {
+                    if ($decision_query->found_posts > 1) {
+                        $doc_count_text = $decision_query->found_posts . ' decisions';
+                    } elseif ($decision_query->found_posts == 1) {
+                        $doc_count_text = '1 decision';
```
MalcolmVonMoJ, line 301: Should be put this in the function that means it would pick up a translation?  
Is it something like `_e("xxx","hale")` - although I think that echoes it out.

MalcolmVonMoJ (dismissed):
It looks like there is a load of CSS which can be dispensed with, and maybe some of the hard-coded text should be put in a translation function - but aside from those minor things, looks good to me.  👍

assets/scss/decisions.scss
brown-a2: Yes, the summary stuff could go. I left some summary code in the listing page with a probably misguided belief that they would be wanted the summary stuff added in. I noticed all our custom post types have their own styles even for page elements that do the same thing. We could probably refactor all of these cpt css classes to use mostly the same ones.

page-decision-listing.php
brown-a2: Yes, probably a good idea.

assets/scss/decisions.scss
MalcolmVonMoJ: Yes, that's probably a good idea.  You could leave this in then and let it be swept away in that refactor work.

PR #285, opened 2024-01-24
Colour relabelling

MalcolmVonMoJ:
Applied the new label wording.
Applied the new colour order.
Removed 3-letter hex codes from defaults.
Blue replaces Red in the list of generic colours.


RobLoweMOJ (approved):
LGTM

PR #286, opened 2024-01-26
Coloured text & remove IE support

MalcolmVonMoJ:
### New text colour settings
Added new colour settings to colour normal page text and headings, and new settings for them on a dark background.

Added new colour settings for links on a dark background - text colour and hover colour only as existing focus styles should be sufficient for all situations.

### Figuring out whether or not a custom background is dark
Added PHP to detect whether a custom background is light or dark and apply style accordingly.
- Said style is held in a new file - brandings-dark-background.scss and duplicated by the PHP when needed.
- Dark background colours are only applied on custom-set dark backgrounds.
- WP custom colour palette doesn't work in IE, so there is no need for the background colours to be in that code block either.

### SVGs
Moved SVGs out of the brandings scss file into their own file, this is then amended to have the relevant colour and added to the CSS file.  SVGs no longer needed in the IE CSS file as they are already IE compatible so no repetition is needed.  

### Internet Explorer support
Removed all IE-specific styles


RobLoweMOJ (commented):
Overall looks like a working solution just want to discuss possible other solutions eg. could the dark background class  be added as part of the pallette options .

Also do we need to add a new JS file. Should we be looking to compress/minify a few of JS files as we got quite. a few now.

RobLoweMOJ:
Dark background text colour not showing in backend

assets/scss/brandings-dark-background--php-used.scss
```diff
@@ -0,0 +1,71 @@
...
+	There are 5 greyscale colours and 4 "generic" colours.
+	The generic colours vary from site to site.
+	They might be dark or they might not be, so we cannot know whether to apply the "dark-text" (etc) colours or not.
+	This file creates the CSS for a dark background, but it uses the class "has-replace_replace-background-color" (which isn't used).
+	We use this CSS in PHP (colour-branding.php) to apply the style should the colour chosen be calculated to be dark.
+	So whilst this CSS is not directly used by anything, it is copied by the PHP, amended and added where needed.
+*******/
+
+@mixin hale-block-inside-block($paletteColour, $colour, $hoverColour: "") {
+	// @include hale-block-inside-block($colour, $hoverColour)
```
RobLoweMOJ, line 12: Is this ment to be commented out? if so put some info why

RobLoweMOJ (approved):
LGTM - one comment

assets/scss/brandings-dark-background--php-used.scss
MalcolmVonMoJ: Removed this line - not needed.

PR #287, opened 2024-01-29
Remove Published list item as stakeholder requested it not needed

brown-a2:
* Remove published list item on Decision CPT listing page.


RobLoweMOJ (approved):
LGTM

PR #288, opened 2024-02-05
Flexible CPTs and Taxonomies

RobLoweMOJ:
- Allow creation of CPTs and Taxonomies in an ACF Option Page
- Creates new listing page where you can set post type, filters and restrictions

### DIFF
--- a/page-listing.php
+++ b/page-listing.php
@@ -41,13 +41,13 @@ while (have_posts()) :
 
         <div class="govuk-grid-row">
             <div class="govuk-grid-column-one-third">
-                <div class="document-listing-search-section">
-                    <div class="document-listing-search-form">
+                <div class="listing-search-section">
+                    <div class="listing-search-form">
                         <form action="<?php echo get_permalink(); ?>" method="GET">
-                            <label for="doc-search-field" class="govuk-visually-hidden">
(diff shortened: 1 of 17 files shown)


page-listing.php
```diff
@@ -0,0 +1,274 @@
...
+                                   value="<?php echo $search_text_HTML; ?>" type="search"
+                                   placeholder="<?php _e('Search', 'hale'); ?>">
+
+                            <?php
+
+                            $listing_filters = get_field('listing_filters');
+
+                            if(!empty($listing_filters) && is_array($listing_filters)) { ?>
+
+                                    <p>Filters</p>
```
brown-a2, line 60: The "Filters" title displaying is based on the $listing_filters being populated in the backend regardless of whether the taxonomy has been assigned to a post or not. Meaning that it is possible to have the Filter title appear with no filter boxes below or posts showing. Probably a post deploy thing to review.

brown-a2 (dismissed):
Besides not liking that we're using the "news" naming thing for the template/object which we've already discussed, everything else is looking good! A few minor comments.

Also, not for this PR but I think at some point we should refactor the page-listing.php to be more modular, but I think that is something to discuss around the whole theme in general.

PR #290, opened 2024-02-27
Quick exit button

RobLoweMOJ:
- The GOVUK Frontend JS is overridden so the ExitPage function opens in the new tab and clears the back button
- Adds new colour options for quick exit button
- Adds option in customiser to turn quick exit button on or off (default off)
- Uses JS to be sticky to top of page on scroll

### DIFF
--- a/assets/js/gov-overrides.js
+++ b/assets/js/gov-overrides.js
@@ -1,0 +1,17 @@
+//Overrides GOVUK exitPage function so it open in newtab
+if (typeof ExitThisPage === 'function') { //check ExitThisPage Class Exists
+  ExitThisPage.prototype.exitPage = function() {
+
+    //Loading Overlay copied from govuk
+    document.body.classList.add("govuk-exit-this-page-hide-content");
+    this.$overlay = document.createElement("div");
+    this.$overlay.className = "govuk-exit-this-page-overlay";
+    this.$overlay.setAttribute("role", "alert");
+    document.body.appendChild(this.$overlay);
+    this.$overlay.textContent = this.i18n.t("activated");
+
+    //Functionality to Open in new Tab
+    window.open("http://bbc.co.uk/weather", "_newtab");
+    window.location.replace("http://www.google.co.uk");  
+  };
+}
--- a/assets/js/hale-scripts.js
+++ b/assets/js/hale-scripts.js
@@ -17,11 +17,14 @@ if( document.getElementById("search-show-hide") != null) {
 
 jQuery( document ).ready(function( $ ) {
 
+
+  /*
   function fleeFromPage(fleeMethod) {
     $("body").hide();
     window.open("http://bbc.co.uk/weather", "_newtab");
     window.location.replace("http://www.google.co.uk");   
   }
+
   $("a.govuk-exit-this-page__button").on("click", function(e) {
     e.preventDefault();
     e.stopPropagation();
@@ -30,18 +33,6 @@ jQuery( document ).ready(function( $ ) {
   $(document).keyup(function(e) {
     if (e.keyCode == 27) { // 27 = escape key
       fleeFromPage("escape key");
-    }
-  });
-
-  /*
-  window.addEventListener('scroll', function() {
-    var stickyDiv = document.getElementById('.govuk-exit-this-page');
-    var stickyDivRect = stickyDiv.getBoundingClientRect();
-  
-    if (window.scrollY >= stickyDivRect.top) {
-      stickyDiv.classList.add('sticky');
-    } else {
-      stickyDiv.classList.remove('sticky');
     }
   });*/
 
--- a/assets/js/hale-scripts.js
+++ b/assets/js/hale-scripts.js
@@ -14,3 +14,80 @@ if( document.getElementById("search-show-hide") != null) {
     }
   );
 }
+
+jQuery( document ).ready(function( $ ) {
+
+  function fleeFromPage(fleeMethod) {
+    $("body").hide();
+    window.open("http://bbc.co.uk/weather", "_newtab");
+    window.location.replace("http://www.google.co.uk");   
+  }
+  $("a.govuk-exit-this-page__button").on("click", function(e) {
+    e.preventDefault();
+    e.stopPropagation();
+    fleeFromPage("mouse click");
+  });
+  $(document).keyup(function(e) {
+    if (e.keyCode == 27) { // 27 = escape key
+      fleeFromPage("escape key");
+    }
+  });
+
+  /*
+  window.addEventListener('scroll', function() {
+    var stickyDiv = document.getElementById('.govuk-exit-this-page');
+    var stickyDivRect = stickyDiv.getBoundingClientRect();
+  
+    if (window.scrollY >= stickyDivRect.top) {
+      stickyDiv.classList.add('sticky');
+    } else {
+      stickyDiv.classList.remove('sticky');
+    }
+  });*/
+
+  $(window).on('scroll resize', function() {
+    var stickyDiv = $('.govuk-exit-this-page');
+    var stickyButton = $('.govuk-exit-this-page a');
+
+    if (stickyDiv.length) {
+
+      var divMinHeight = stickyButton.outerHeight();
+      var divMinWidth = stickyButton.outerWidth();
+
+      if ($(".hale-header__mobile-controls--search").is(":visible")) {
+        //divMinHeight = 'initial';
+        divMinWidth = 0
+      }
+
+      stickyDiv.css({
+        "min-height": divMinHeight,
+        "min-width": divMinWidth
+       });
+
+      var stickyDivTop = stickyDiv.offset().top;
+      var stickyDivLeft = stickyDiv.offset().left;
+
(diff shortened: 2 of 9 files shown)


inc/colours.php
```diff
@@ -111,6 +111,17 @@ function hale_get_colours() {
 			['button-focus-text',$black,'Button text when focused','',''],
 			['button-focus-outline',$yellow,'Button outline when focused (including search button)','This is thicker than any set button border',''],
 
+			//Quick Exit button
+			['quick-exit-button',$green,'Quick Exit Button background','',''],
```
MalcolmVonMoJ, line 115: The default should probably be `$red`, rather than `$green`

inc/colours.php
```diff
@@ -111,6 +111,17 @@ function hale_get_colours() {
...
+			['quick-exit-button-text',$white,'Quick Exit Button text','',''],
+			['quick-exit-button-border',"none",'Quick Exit Button border','Enter a complete CSS value for border, e.g. "solid 2px #0b0c0c", "none"','text'], // not a colour
+			['quick-exit-button-hover',$buttonHoverGreen,'Quick Exit Button background on hover','',''],
```
MalcolmVonMoJ, line 118: The hover colour should probably be `$darkRed`

MalcolmVonMoJ, line 118:
```suggestion
			['quick-exit-button-hover',$darkRed,'Quick Exit Button background on hover','',''],
```

inc/colours.php
MalcolmVonMoJ, line 115:
```suggestion
			['quick-exit-button',$red,'Quick Exit Button background','',''],
```

assets/js/gov-overrides.js
```diff
@@ -0,0 +1,17 @@
...
+    this.$overlay.className = "govuk-exit-this-page-overlay";
+    this.$overlay.setAttribute("role", "alert");
+    document.body.appendChild(this.$overlay);
+    this.$overlay.textContent = this.i18n.t("activated");
+
+    //Functionality to Open in new Tab
+    window.open("http://bbc.co.uk/weather", "_newtab");
+    window.location.replace("http://www.google.co.uk");  
+  };
+}
```
MalcolmVonMoJ, line 17: Can we also add in a function override for the shift keys to get them to behave in the same way?  

The current GDS function is 
```
  ExitThisPage.prototype.handleKeypress = function (event) {
    // Detect if the 'Shift' key has been pressed. We want to only do things if it
    // was pressed by itself and not in a combination with another key—so we keep
    // track of whether the preceding keyup had shiftKey: true on it, and if it
    // did, we ignore the next Shift keyup event.
    //
    // This works because using Shift as a modifier key (e.g. pressing Shift + A)
    // will fire TWO keyup events, one for A (with e.shiftKey: true) and the other
    // for Shift (with e.shiftKey: false).
    if (
      (event.key === 'Shift' || event.keyCode === 16 || event.which === 16) &&
      !this.lastKeyWasModified
    ) {
      this.keypressCounter += 1

      // Update the indicator before the below if statement can reset it back to 0
      this.updateIndicator()

      // Clear the timeout for the keypress timeout message clearing itself
      if (this.timeoutMessageId !== null) {
        clearTimeout(this.timeoutMessageId)
        this.timeoutMessageId = null
      }

      if (this.keypressCounter >= 3) {
        this.keypressCounter = 0

        if (this.keypressTimeoutId !== null) {
          clearTimeout(this.keypressTimeoutId)
          this.keypressTimeoutId = null
        }

        this.exitPage()
      } else {
        if (this.keypressCounter === 1) {
          this.$updateSpan.innerText = this.i18n.t('pressTwoMoreTimes')
        } else {
          this.$updateSpan.innerText = this.i18n.t('pressOneMoreTime')
        }
      }

      this.setKeypressTimer()
    } else if (this.keypressTimeoutId !== null) {
      // If the user pressed any key other than 'Shift', after having pressed
      // 'Shift' and activating the timer, stop and reset the timer.
      this.resetKeypressTimer()
    }

    // Keep track of whether the Shift modifier key was held during this keypress
    this.lastKeyWasModified = event.shiftKey
  }

}
```

languages/cy.po
```diff
@@ -19990,3 +19990,15 @@ msgstr "&copy; Hawlfraint"
 #: hale/partials/footer-licence.php
 msgid "All content is available under the <a class=\"govuk-footer__link\" href=\"%s\" rel=\"license\">Open Government Licence v3.0</a>, except where otherwise stated"
 msgstr "Mae’r holl gynnwys ar gael o dan <a class=\"govuk-footer__link\" href=\"%s\" rel=\"license\">Drwydded Llywodraeth Agored v3.0</a>, ac eithrio lle nodir yn wahanol"
+
```
MalcolmVonMoJ, line 19993: I think we need more translations.  These are the Welsh from the GDS PR:
```
      text: Gadael y dudalen
      activatedText: Tudalen ymadael
      timedOutText: Wedi’i amseru
      pressTwoMoreTimesText: Pwyswch 'Shift' 2 gwaith arall
      pressOneMoreTimeText: Pwyswch 'Shift' 1 mwy o amser
```
The last three texts are visually hidden texts for screen readers - they will be announced when shift keys are pressed.  

Also note how the translations put Shift in quotes, maybe that is a good idea.

assets/js/gov-overrides.js
RobLoweMOJ: This override already works for the shift keys aswell. if you see in the function it calls this.exitPage()

MalcolmVonMoJ (dismissed):
I think this is good to go.  
I have recommended a few small changes, and there might be places we want to improve later.  But this works well and I think is fully suitable for deployment to live, especially as stakeholders seem to want it now.

assets/js/gov-overrides.js
MalcolmVonMoJ: It doesn't seem to work - at least not for me locally.  I looked at the code and expected it to work but it didn't.  Is it working for you?  
(Quick Exit worked, just didn't open the other tab.)

RobLoweMOJ: It did for me. Are you sure your popup blocker isnt stopping it..It did for me on local the first time.

languages/cy.po
```diff
@@ -19990,3 +19990,15 @@ msgstr "&copy; Hawlfraint"
...
+#: hale/header.php
+msgid "Exit this page"
+msgstr "Gadael y dudalen hon"
```
MalcolmVonMoJ, line 19996:
```suggestion
msgstr "Gadael y dudalen"
```

inc/colours.php
RobLoweMOJ: Good idea changed

inc/customizer.php
```diff
@@ -130,6 +130,32 @@ function hale_customize_register( $wp_customize ) {
...
+		array(
+			'default'           => 'no',
+			'sanitize_callback' => 'hale_sanitize_select',
+		)
+	);
+	$wp_customize->add_control(
+		'show_quick_exit',
+		array(
+			'label'       => esc_html__( 'Show Quick Exit Button?', 'hale' ),
+			'description' => esc_html__( 'Would you like to quick exit button in the header?', 'hale' ),
```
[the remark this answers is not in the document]
RobLoweMOJ: Fixed

inc/colours.php
RobLoweMOJ: changed as suggested

MalcolmVonMoJ (approved):
Looks good to me

assets/js/hale-scripts.js
```diff
@@ -14,3 +14,54 @@ if( document.getElementById("search-show-hide") != null) {
...
+
+  //Keeps the Quick Exit Button at top of page once scrolled past the buttons original position in header
+  $(window).on('scroll resize', function() {
+    var stickyDiv = $('.govuk-exit-this-page');
+    var stickyButton = $('.govuk-exit-this-page a');
+
+    if (stickyDiv.length) {
+
+      //Keeps the surrounding Div the same width and height even when the button is fixed position. As this is used to work out the buttons left pos.
+      var divMinHeight = stickyButton.outerHeight();
```
brown-a2, line 28: Is it worth adding the unit as well, so like  var divMinHeight = stickyButton.outerHeight() + 'px';

brown-a2, line 28: Is it worth adding in the unit as well?  var divMinHeight = stickyButton.outerHeight() + 'px';

assets/js/hale-scripts.js
```diff
@@ -14,3 +14,54 @@ if( document.getElementById("search-show-hide") != null) {
     }
   );
 }
+
+jQuery( document ).ready(function( $ ) {
+
+  //Keeps the Quick Exit Button at top of page once scrolled past the buttons original position in header
+  $(window).on('scroll resize', function() {
+    var stickyDiv = $('.govuk-exit-this-page');
+    var stickyButton = $('.govuk-exit-this-page a');
```
brown-a2, line 23: Is it easier to add the one to the other like 
```
var stickyDiv = $('.govuk-exit-this-page');
var stickyButton = stickyDiv.find('a');
```

brown-a2 (approved):

PR #291, opened 2024-02-27
Changed colour input back to text from colour picker

MalcolmVonMoJ:
Revert to a text entry for colours in customizer.  
Note added to aid changing back to a colour picker should this be needed.

NB: 

If doing bulk-entry and many changes, text entry is far easier to work with.
If reviewing current colours, the colour-picker preview is far more useful.  
Therefore we might need to change this back as per the main needs of the task/individual dealing with it.


RobLoweMOJ (dismissed):
LGTM

PR #296, opened 2024-03-11
GDS Design System upgrade to v5.2

MalcolmVonMoJ:
Upgrades to the latest GDS design system.  
No real impact for the Hale theme.
Font sizes have changed - but we override these styles with our own.
New crown has finally been introduced (only 5 months after we were told to use it), but we have already made this change where needed - and our own local code overrides this.


MalcolmVonMoJ:
Superseded by #310

PR #302, opened 2024-03-15
Banners

MalcolmVonMoJ:
- Emergency banner
  - Choices for Mourning/National/Local emergency.
- Critical information banner
  - Javascript to remove info banner on click of "hide this message"
- Added cookie to remember that info banner has been hidden.
  - Cookie hasn't got an expiry date, will expire on browser closing.
  - Cookie remembers banner content - if content changes, the banner will reappear.
- New section in site dashboard - **Site Customization** which has two separate subsections.
  - Critical information banner
  - Emergency banner

__On roll-out, every site will need to have `info_banner_dismissed` added to their essential cookie list in CCfW__
Cookie Compliance For WordPress > Cookie Management > Essential cookies > Add cookie
Cookie name must be
 `info_banner_dismissed`
Information needs to say something like: 
`This remembers whether an information banner has been dismissed to prevent it from being shewn again.`
Expiry must be:
`When you close your browser`


header.php
```diff
@@ -122,6 +125,10 @@ function wp_body_open() {
 </div>
 </header>
 <?php
+
+include "inc/information-banner.php";
```
RobLoweMOJ, line 129: Please could we move this file to partials directory with other page elements

inc/emergency-banner.php
```diff
@@ -2,61 +2,87 @@
 
 if( function_exists('get_field') ) {
```
RobLoweMOJ, line 3: Please could we move this file to partials directory with other page elements

inc/banner-settings.php
```diff
@@ -0,0 +1,14 @@
+<?php
+if( function_exists('acf_add_options_page') ) {
```
RobLoweMOJ, line 2: As previously discussed not a blocker but this cna now be done using the JSON Options Pages - Ideally we would want to standardsise the approach but we can have a dev discussion about the pros and cons.

functions.php
```diff
@@ -252,6 +252,7 @@ function hale_scripts() {
     wp_enqueue_script('hale-scripts', hale_mix_asset('/js/hale-scripts.js'), '', null, true);
     wp_enqueue_script('hale-skip-link-focus-fix', hale_mix_asset('/js/skip-link-focus-fix.js'), '', null, true);
     wp_enqueue_script('hale-navigation', hale_mix_asset('/js/navigation.js'), '', null, true);
+    wp_enqueue_script('hale-close-button', hale_mix_asset('/js/banner-close-button.js'), '', null, true);
```
RobLoweMOJ, line 255: Does this need to be a seperate JS file? Could it be minified with other scripts. Same with hale-skip-link-focus-fix and hale-navigation

acf-json/group_65f00eaa4af50.json
```diff
@@ -0,0 +1,408 @@
...
+                "id": ""
+            },
+            "default_value": "National emergency",
+            "maxlength": 50,
+            "placeholder": "",
+            "prepend": "",
+            "append": ""
+        },
+        {
+            "key": "field_65f013020e0c1",
```
RobLoweMOJ, line 243: Do we need seperate title fields or could we just have one used by all?

inc/emergency-banner.php
```diff
@@ -2,61 +2,87 @@
...
+        case "No banner":
+            $active = false;
+            break;
+        case "National emergency":
+            $active = true;
+            $banner_type_class .= "--national-emergency";
+            $banner_title = get_field('banner_national_emergency_title', 'option');
+            $banner_subtext = get_field('banner_summary', 'option');
+            if (is_null($banner_title) || trim($banner_title) == "" ) {
+                if (is_null($banner_subtext) || trim($banner_subtext) == "" ) {
```
RobLoweMOJ, line 21: You have made the summary required here and yet the field isnt..should this be required?

inc/emergency-banner.php
MalcolmVonMoJ: Done

header.php
MalcolmVonMoJ: Done

acf-json/group_65f00eaa4af50.json
MalcolmVonMoJ: I've made them separate fields because I wanted the hint text to be different for local and national.  I couldn't see a way of doing that otherwise.  Summary is shared by local and national.

inc/emergency-banner.php
MalcolmVonMoJ: This is a "nothing entered" check.  If both title and summary are blank, then the banner is blank.  This should not happen, but might if someone enters a space for each field.

The title can revert to the default title if only that is blank, but if both are blank, we remove the banner.  

I have added a safeguard against the default being left in the title but no summary, which is more likely than the above.

functions.php
MalcolmVonMoJ: Combined them and reduced enqueing.

RobLoweMOJ (approved):
LGTM

PR #306, opened 2024-03-20
Colour sections

MalcolmVonMoJ:
Split the different colours out into sections.

In the main function:
- Added a new (associative) dimension to the master colour array.
- Added a new argument for the function (`$sections`).
- Existing colours are amalgamated if `$sections` is false (or missing)

In the customizer:
- Added a new panel to replace the `colors` section.
- Several new sections for the colours.
- Moved the GDS colours and stuff to the top section.  
- New control in the top section, where a user can switch from text entry to colour picker.
- Reworded "coloured bar" to "departmental branding" - as this colour might be used elsewhere.


RobLoweMOJ (approved):
LGTM. Double check colours are good on staging with a few of the sites (when deployed)

PR #310, opened 2024-03-28
Heading sizes

MalcolmVonMoJ:
- Moved to using GDS heading sizes and font sizes.
  - Removed the defunct `hale-heading-x` sizes.  
    - M and XXL were not used.  
    - S was already the same size as GDS M - so this replaces it.  
    - L was nearest to GDS L, XL was nearest to GDS XL so replaced like that.  
  - Made the `has-x-font-size` classes use GDS sizes.
    - New CSS variables to make this work within the existing WP methodology.
- Adopted the latest GDS mixins.
  - New mixins from GDS Design System v5.0.2, so this latest version has also been brought in in this PR.
- CSS rationalisation, reduce superfluous rules.
- Added font size options to list, list items and paragraphs to fix existing bug.
- Removed drop cap option from paragraph.

### DIFF
--- a/assets/scss/brandings-extended-core-blocks.scss
+++ b/assets/scss/brandings-extended-core-blocks.scss
@@ -1,4 +1,4 @@
-html.hale-colours-variable {
+body.hale-colours-variable {
 	//Group Block
 	.wp-block-group:not(.has-background) {
 		&.is-style-bleeding-background,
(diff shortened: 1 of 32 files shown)


RobLoweMOJ (approved):
LGTM though as discussed still may be worth looking into if we can get rid of font global styles that we having to ovverride.

As also discussed today. We may need to sync this with the framework updates on blocks and cookie banner. Lets discuss tomoz.

PR #314, opened 2024-04-11
Added template class to editor and applied GDS width classes

MalcolmVonMoJ:
Applied the GDS widths to the editor screen by extending the GDS classes.  
- `govuk-grid-column-full-from-desktop` for full width
- `govuk-grid-column-two-thirds` for default

Added a new bit of PHP to add a class so the editor knows which template is being used.

Added lists to the font size declaration, so the editor has the same font size as the frontend for lists.

### DIFF
--- a/inc/template-functions.php
+++ b/inc/template-functions.php
@@ -168,15 +168,20 @@ function hale_get_branding_class() {
  * Get the template which has been selected for the page
  */
 function hale_get_template_class() {
-	$template = get_page_template_slug();
-	if ($template) {
-		$class = preg_replace('/\.php$/', "", $template);
-		$classes = " hale-template hale-template--" . sanitize_html_class($class);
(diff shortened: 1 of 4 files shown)


inc/template-functions.php
```diff
@@ -164,6 +164,23 @@ function hale_get_branding_class() {
 	return $style_class;
 }
 
+/**
+ * Get the template which has been selected for the page
+ */
+function hale_get_template_class() {
+	$template = get_page_template_slug();
```
RobLoweMOJ, line 171: This will run on all admin pages. Please make it more specific by checking get_current_screen. Thanks

MalcolmVonMoJ: Added

RobLoweMOJ (approved):
LGTM

PR #322, opened 2024-05-09
GDS colours overwritten bug

MalcolmVonMoJ:
Some "GDS" colours were being overwritten by some CSS added elsewhere.  

This was affecting the staggered box on Magistrates Recruitment.

This increases the specificity of the GDS colour CSS to match that of the colour variable declarations.  The two should now be in synch.


RobLoweMOJ (approved):
LGTM

PR #325, opened 2024-05-23
Dash update

brown-a2:
* Update theme style sheet to latest changes and add licence info
* Remove metaboxes to keep the dashboard a bit more clean for editors.
- Boxes removed, wp news, yoast news, and wp health check which doesn't work correctly with our platform config


RobLoweMOJ (approved):
LGTM

PR #326, opened 2024-05-23
Add check on JSON data returned fix array null error

brown-a2:
Two bugs fixed here:
* One relating to screen function receiving NULL when it cannot. 
* Second, was on a page, an array not checked when retrieving JSON.


RobLoweMOJ (approved):
LGTM

PR #329, opened 2024-05-28
Added new colours to colour picker

MalcolmVonMoJ:
- Adds new colour options for the info banner
- Also adds missing text colour options for the header search box


RobLoweMOJ (approved):
LGTM

PR #330, opened 2024-05-29
Colour bugs

MalcolmVonMoJ:
- Made colour hex code case-insensitive for contrast check
- Pulled through the correct hover colour for links on a dark background
- Set a background colour for the HTML tag - so the footer background colour extends to the bottom on short pages


RobLoweMOJ (approved):
LGTM

PR #335, opened 2024-06-05
Editor page width

MalcolmVonMoJ:
Restored the styles which limit the width of the preview in the editor pane.  But did so in a way which prevents the styles from leaking through to the backend editor controls.  

Taken the opportunity to reduce the repetition in the CSS as detailed below.  Much CSS was called twice by having the same code `@import`ed twice, or by having one file used for both frontend and editor backend.  

### CSS re-organisation:
- `custom-branding.scss`
  - Removed duplication (`brandings-mojblocks` was called twice)
  - Removed the call to `editor.scss`, as this was being called by the frontend as well
  - This is still a general file mainly used to compile the colour branding files - called by both frontend and editor
- `editor-branding` **new file**
  - Enqueued when in editor view (in `inc/custom-gutenberg.php`), calls the below `editor.scss` and the above `custom-branding.scss` 
- `editor-backend-controls.scss` **file deleted**
- `editor.scss`
  - The call to the fonts CSS file is now within a class to prevent the font styles leaking out into the other editor controls in the sidebar (the shouty bits)
  - Deleted reference to the above `editor-backend-controls.scss`
  - Added in calls to other stylesheets (`default.scss` and `hale-styles.scss`) which were previously in the above `custom-branding.scss`.  This reduces CSS repetition.
  - Removes the `template-mismatch` class which warned if a template had been selected but not refreshed.  This ceased working in WP 6.5.3.
  - Added a background colour (white) so there isn't a grey bar for the right two-thirds of the default template view.
- `pt-sans.scss`
  - Removed the `@extend` of the GDS heading classes and replaced with the relevant font sizes.

### Other changes:
- Amended the bit of code in template function which adds a class to the body tag of the page.
  - Previously this was never visible as it was overwritten (`=` instead of `.=`).  
  - Renamed the class to conform with CSS naming convention.


RobLoweMOJ:
Im still getting the styles bleed when i test this locally. I think we need to target edit-post-visual-editor or editor-styles-wrapper

RobLoweMOJ (approved):
LGTM

PR #337, opened 2024-06-07
Colour bug

MalcolmVonMoJ:
Adding in the HTML background colour destroyed some of the other styling, so we revert that change here.


RobLoweMOJ (approved):
LGTM

PR #338, opened 2024-06-07
Taxonomies and Published Dates

RobLoweMOJ:
Listing Page: Adds option to display fields so you can display taxonomies

CPT settings: 

- Add display settings - advanced tab
- Toggle Published Date on/off on single view
- Toggle Taxonomies on/off on single view
- Dropdown select to decide in taxonomies show as list or tags

Side note:

We havent merged the docs single and flex single template yet as this would require a show summary toggle which is part of the news migration

We havent added the show_published_date_on_single_view on the doc single as we first need to roll out the published date toggle.

### DIFF
--- a/template-parts/flexible-cpts/details.php
+++ b/template-parts/flexible-cpts/details.php
@@ -11,29 +11,33 @@
     <?php 
     
     $groups = acf_get_field_groups(array('post_type' => $post_type)); 
-    
+
     if(is_array($groups) && count($groups) > 0){
 
         foreach($groups as $group) {
 
             $fields = acf_get_fields($group['key']);
 
-            if(is_array($fields) && count($fields) > 0){
-
-                foreach($fields as $field){
-                    
-                    $field_value = get_field($field['name']);
-
-                    if(!empty($field_value) && $field["single_view"]){
-                        ?>
-                            <div class="flexible-post-type-detail">
-                                <div class="flexible-post-type-detail-label"><?php echo $field['label']; ?>: </div>
-                                <?php echo $field_value; ?>
-                            </div>
-                        <?php
-                    }
-                }
+            if (empty($fields)) {
+                continue;
             }
+
+            foreach($fields as $field){
+                
+                $field_value = get_field($field['name']);
+
+                if(empty($field_value) || !array_key_exists('single_view',  $field) || !$field["single_view"]){
+                    continue;
+                }
+                ?>
+                    <div class="flexible-post-type-detail">
+                        <div class="flexible-post-type-detail-label"><?php echo $field['label']; ?>: </div>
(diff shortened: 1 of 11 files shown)


template-parts/flexible-cpts/details.php
```diff
@@ -0,0 +1,122 @@
...
+
+            $fields = acf_get_fields($group['key']);
+
+            if(is_array($fields) && count($fields) > 0){
+
+                foreach($fields as $field){
+                    
+                    $field_value = get_field($field['name']);
+
+                    if(!empty($field_value) && $field["single_view"]){
```
brown-a2, line 27: I get an error message on this page "Undefined array key "single_view"".  When I look at $field["single_view"]  a get NULL. I can wrap a !empty() around $field["single_view"] to quiet it but that's kind of hidding the fact it is coming back null.

template-parts/flexible-cpts/details.php
```diff
@@ -0,0 +1,122 @@
...
+ $post_type = get_post_type();
+ 
+?>
+
+<div class="flexible-post-type-details">
+    <?php 
+    
+    $groups = acf_get_field_groups(array('post_type' => $post_type)); 
+    
+    if(is_array($groups) && count($groups) > 0){
```
brown-a2, line 15: If you didn't want so many nested ifs, you could do something like

```
    if (empty($groups)) {
        return;
    }
    
    foreach ($groups as $group) {
        $fields = acf_get_fields($group['key']);
    
        if (empty($fields)) {
            continue;
        }
    
        foreach ($fields as $field) {
            $field_value = get_field($field['name']);
    
            if (empty($field_value) || empty($field["single_view"])) {
                continue;
            }
            ?>
            <div class="flexible-post-type-detail">
                <div class="flexible-post-type-detail-label"><?php echo esc_html($field['label']); ?>: </div>
                <?php echo esc_html($field_value); ?>
            </div>
            <?php
        }
    }  
```

template-parts/flexible-cpts/details.php
```diff
@@ -0,0 +1,122 @@
...
+
+                foreach($fields as $field){
+                    
+                    $field_value = get_field($field['name']);
+
+                    if(!empty($field_value) && $field["single_view"]){
+                        ?>
+                            <div class="flexible-post-type-detail">
+                                <div class="flexible-post-type-detail-label"><?php echo $field['label']; ?>: </div>
+                                <?php echo $field_value; ?>
```
brown-a2, line 31: Not sure if worth escaping? ```<?php echo esc_html($field_value); ?>```

brown-a2 (commented):
Only major issue is I'm getting a error , undefined array key. Nothing else major. Once we've got that sorted looks good to go.

brown-a2 (approved):

PR #344, opened 2024-06-17
Add feature image support to single post type page

brown-a2:
* Adds an image to the single view page if an image is used.
* Added to both documents and flex single pages.


MalcolmVonMoJ (dismissed):
This looks good to me

PR #348, opened 2024-06-18
Show Summary - on post

RobLoweMOJ:


MalcolmVonMoJ (approved):
Looks good to me

inc/acf/post-type-fields/field-summary.php
```diff
@@ -18,6 +18,28 @@ function hale_register_post_type_fields_summary()
         if (isset($post_type->post_summary) && $post_type->post_summary == '1') {
             $post_type_name = $post_type->label;
             $post_type_key = $key;
+            
+            //Only show field if sitewide show summary setting is on and has single view
+            if (isset($post_type->publicly_queryable) && $post_type->publicly_queryable == true && isset($post_type->show_summary_on_single_view) && $post_type->show_summary_on_single_view == '1') {
+                // Add the ACF field for the show summary
```
brown-a2, line 24: Why do we have to check if publicly queryable, if single view is turned off it doesn't matter if summary is on or off does it?

PR #359, opened 2024-07-23
Breadcrumbs overflow

MalcolmVonMoJ:
Increased max breadcrumb width to 100% of the breadcrumb list (full page width).
Clipped the overflow so it doesn't run off the page.


RobLoweMOJ (approved):
LGTM

PR #361, opened 2024-07-30
Header search box re-hash

MalcolmVonMoJ:
- Changed HTML for the two search boxes so they are different, pass through an attribute into the function when it is in the header.
- Removed rounded corners
- Added new colours and borders to create a seamless look between search box and search button.


RobLoweMOJ (approved):
LGTM

PR #365, opened 2024-08-07
Tag accessibility

MalcolmVonMoJ:
Added visually hidden text to tags and categories.  
Enclosed tag/category text in `<strong>`.
No visual or behaviour changes.


brown-a2 (approved):
Do we need to sanitize the output here so something like ```<?php echo esc_html($story_category->name); ?>``` for both?

PR #366, opened 2024-08-07
Empty pagination buttons

MalcolmVonMoJ:
Added a bit of styling so empty buttons are `display:none`.
Removed line returns in the code so the above approach works.


RobLoweMOJ (approved):
LGTM

PR #367, opened 2024-08-08
More button focus

MalcolmVonMoJ:
Changed the CSS to allow the more button's content to be visible when the more button is tabbed to.  
- Changed `display:none` to `visibility:hidden`, so a transition can be used.
- Transition used to ensure there is a small delay in hiding the menu, this delay is long enough so the focus can be successfully transferred to the menu hidden by the more button.


RobLoweMOJ (approved):
LGTM

PR #369, opened 2024-08-28
Colour warning emails

MalcolmVonMoJ:
Checked for unset colours
- if any colour is unset and on prod, a "notice" is issued
- if no colours are set at all and on prod, a "warning" is issued
- if no colours are set and not on prod, a "notice" is issued

(Notices and warnings here are all emails, just prefixed differently.)


RobLoweMOJ (dismissed):
LGTM

PR #383, opened 2024-09-26
Featured item block and API extension

MalcolmVonMoJ:
Moved the API extension code from MoJ Blocks to Hale

### DIFF
--- a/inc/api-extensions.php
+++ b/inc/api-extensions.php
@@ -1,16 +1,22 @@
 <?php
-function hale_extend_wp_api($data, $post, $context) {
+function hale_extend_wp_api_image($data, $post, $context) {
 
     // Add featured image to API
-    $featured_image_id = $data->data['featured_media']; // get featured image id
-    $featured_image_url = wp_get_attachment_image_src( $featured_image_id, 'feature' ); // get url of the feature size
+    $featured_image_url = false;
+    if (array_key_exists('featured_media',$data->data)) {
+        $featured_image_id = $data->data['featured_media']; // get featured image id
+        $featured_image_url = wp_get_attachment_image_src( $featured_image_id, 'feature' ); // get url of the feature size
+    }
 
     if( $featured_image_url ) {
         $data->data['featured_image_url'] = $featured_image_url[0];
     } else {
         $data->data['featured_image_url'] = "";
     }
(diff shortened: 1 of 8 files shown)


inc/api-extensions.php
```diff
@@ -0,0 +1,30 @@
+<?php
+function hale_extend_wp_api($data, $post, $context) {
+
+    // Add featured image to API
+    $featured_image_id = $data->data['featured_media']; // get featured image id
```
RobLoweMOJ, line 5: What happens if featured image is off? No array key?

inc/api-extensions.php
```diff
@@ -0,0 +1,30 @@
...
+    $featured_image_url = wp_get_attachment_image_src( $featured_image_id, 'feature' ); // get url of the feature size
+
```
RobLoweMOJ, line 7: what happens if $featured_image_id is blank

inc/api-extensions.php
```diff
@@ -0,0 +1,30 @@
...
+    return $data;
+}
+
+function hale_add_all_prepare_filters() {
+    $args = array(
+        'public'   => true,
+    );
+    $post_types = get_post_types($args);
+    foreach ($post_types as $post_type) {
+        add_filter( "rest_prepare_$post_type", 'hale_extend_wp_api', 10, 3 );
```
RobLoweMOJ, line 27: do we check if features are enabled?

inc/api-extensions.php
MalcolmVonMoJ: If it is blank/null/false, then it will be set to a zero-length string by this code:
```
    if( $featured_image_url ) {
        $data->data['featured_image_url'] = $featured_image_url[0];
    } else {
        $data->data['featured_image_url'] = "";
    }

inc/api-extensions.php
MalcolmVonMoJ: Added `array_key_exists` to check for that.

inc/api-extensions.php
MalcolmVonMoJ: I'm now setting this to false to begin with, then checking the array key exists as above.

RobLoweMOJ (approved):
LGTM

PR #386, opened 2024-10-07
Submenu background

MalcolmVonMoJ:
When the custom colour failed (hopefully fixed in the previous PR), the most serious issue was the background colour for submenus being transparent.  

This adds a fallback to keep the background black (same as header).


RobLoweMOJ (approved):
LGTM

PR #387, opened 2024-10-07
Adopted GDS next and previous buttons

MalcolmVonMoJ:


inc/pagination.php
```diff
@@ -42,12 +42,21 @@ function hale_archive_pagination($template = '', $custom_query = '')
                     <?php
                         if ($current_page_number > "1") {
                             echo "<li class='archive-pagination-prev-btn'>";
-                            previous_posts_link('< Previous', $max_pages);
+                            previous_posts_link('
+                            <svg class="govuk-pagination__icon govuk-pagination__icon--prev" xmlns="http://www.w3.org/2000/svg" height="13" width="15" aria-hidden="true" focusable="false" viewBox="0 0 15 13">
+                                <path d="m6.5938-0.0078125-6.7266 6.7266 6.7441 6.4062 1.377-1.449-4.1856-3.9768h12.896v-2h-12.984l4.2931-4.293-1.414-1.414z"></path>
+                            </svg><span class="govuk-pagination__link-title">Previous<span class="govuk-visually-hidden"> page</span>
```
RobLoweMOJ, line 48: Does the previous text need to be translated to welsh?

MalcolmVonMoJ: Added a translation tag around the text.  
Note that there is a visually hidden suffix which will also need to be translated:
`Previous<span class="govuk-visually-hidden"> page</span>`
I have wrapped all of this in the translation tag - this is because a language might put the "page" bit before the "previous" bit.  It does mean that there is HTML in the translation, but I don't really see a big issue with that.

brown-a2:
I could be wrong but doesn't previous_posts_link() and next_posts_link() handle localization/translation already?

MalcolmVonMoJ:
I just had a go and it doesn't seem to call the translation.  
Even if it did, I'm not sure we want to have all that SVG code in the translation file.  

As part of testing that, I added the Welsh translations (from the Welsh Senate's website) so I have added them into this PR too.

RobLoweMOJ (approved):
LGTM

PR #394, opened 2024-10-15
Add Fix for pagination causing link overlay

RobLoweMOJ:


brown-a2 (approved):

MalcolmVonMoJ:
The `:after` style seems to be to slightly increase the click area for the button.  Our HTML structure is slightly different and as such there was no `position:relative` style applied.  So the `:after`'s `position:absolute` was absolute in relation to the whole page.  

All links in the top section of the page were affected, this resulted in a click trap, where header links would just have mimicked the next/previous link.  

This fix is fine.  We could put the `position:relative` on the class `archive-pagination-prev-btn`, but it would achieve nothing different.  Our HTML structure means that the click area would not be increased by this.  

I am satisfied with this solution in the long-term, with the slight caveat that our click areas are smaller than those of GDS.  As there is an accessibility review happening on PPJ, which includes this pagination, we can pick the click area up with them then.

PR #397, opened 2024-10-23
Added H1 check

MalcolmVonMoJ:
On homepage only, checks for existing H1, if none exists, adds one in.

### DIFF
--- a/page-full-width.php
+++ b/page-full-width.php
@@ -52,7 +52,7 @@ while (have_posts()) :
         ?>
         </div>
     <?php
-    } elseif (is_front_page() !== false) {
+    } elseif (is_front_page()) {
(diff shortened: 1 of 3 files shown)


page-full-width.php
```diff
@@ -52,6 +52,13 @@
         ?>
         </div>
     <?php
+    } elseif (is_front_page() !== false) {
```
RobLoweMOJ, line 55: Could simplify to elseif (is_front_page()) as function only returns true or false

RobLoweMOJ (dismissed):
LGTM

page-full-width.php
MalcolmVonMoJ: Changed 👍

PR #399, opened 2024-10-29
Restrict Blocks - Add Pattern and Custom HTML Blocks. 

RobLoweMOJ:
Add Pattern and Custom HTML Blocks.
Remove old featured document block.


RobLoweMOJ:
the item list block is not live yet

PR #401, opened 2024-11-04
List background bug

MalcolmVonMoJ:
Corrected a padding issue when a list has both a width style (e.g.`is-style-default`) and a coloured background.


RobLoweMOJ (approved):
LGTM

PR #409, opened 2025-01-14
Introduce hearing list template and supporting features

brown-a2:
This new WP template provides a dedicated list template style page for hearings. We've built this with a focus on the OBI requirements but have designed it to be flexible for use on future hearings. It does not currently allow for subtopics. As it was designed with OBI in mind some taxonomies are hardcoded into it. Witnesses being one.

Other features it introduces
- Autocomplete field
- Date range fields

## Checklist

- [ x] Checked on a mobile device
- [x ] Checked on a desktop device
- [ ] Checked on Safari
- [x ] Checked on Firefox
- [ x] Checked on Chrome


brown-a2:
This has been reviewed by all devs.

PR #410, opened 2025-01-16
Fix border appearing around non-hearing tags

brown-a2:
Fixes a border appearing around non-hearing tags. CSS had been put outside the hearing parent class so it became globally scoped.


MalcolmVonMoJ (approved):
LGTM

PR #411, opened 2025-01-20
No witness bug

MalcolmVonMoJ:

Fixes a bug ion the HTML which occurred when no witnesses were on any hearing.

4.18.2

Neither.

## Checklist

- [ ] Checked on Safari
- [ ] Checked on Firefox
- [x] Checked on Chrome


RobLoweMOJ (approved):
LGTM

PR #417, opened 2025-02-18
Printing headers and footers

MalcolmVonMoJ:

This makes headers and footers more printer-friendly
- all navigation/search boxes are removed
- links in the footer are removed, except copyright links (often a logo), emails and telephone numbers
- new printer friendly logo is now available
  - if not selected, the old logo will use the header background colour
- header and footer backgrounds are white

Neither.

## Checklist

- [ ] Checked on Safari
- [ ] Checked on Firefox
- [x] Checked on Chrome


footer.php
```diff
@@ -83,6 +83,42 @@ class="govuk-footer__link govuk-footer__copyright-logo"
...
+        get_template_part( 'partials/footer-licence' );
+        if ('yes' != $crown_copyright) {
+          get_template_part( 'partials/footer-copyright' );
+        }
+      ?>
+    </div>
+    <?php
+      if ('yes' == $crown_copyright) {
+        $copyright_style = "padding-top:0; background-image:none;";
+    ?>
```
RobLoweMOJ, line 99: I would recommend making crown copyright a template part

footer.php
```diff
@@ -83,6 +83,42 @@ class="govuk-footer__link govuk-footer__copyright-logo"
...
+      <div class="govuk-footer__meta-item">
+        <a
+          class="govuk-footer__link govuk-footer__copyright-logo"
+          href="https://www.nationalarchives.gov.uk/information-management/re-using-public-sector-information/uk-government-licensing-framework/crown-copyright/"
```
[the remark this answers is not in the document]
MalcolmVonMoJ: I have reverted to one.

footer.php
MalcolmVonMoJ: Done

MalcolmVonMoJ:
Yes, well, probably - I'll see what GDS do again, but probably yes.  

I'm thinking I need to hide the language switch too... 🤔

MalcolmVonMoJ:
I've made changes, but a colour refresh will be needed as the PHP generated CSS file has changed and this is only changed when a colour is changed.  Only needed for MagiRec as that is the only external site with GDS styles at the moment.

RobLoweMOJ (approved):
LGTM

PR #418, opened 2025-02-19
Sticky table of contents

MalcolmVonMoJ:

Makes the table of contents for CPTs sticky:
- The code for the former horizontal scroll issue has had to be altered slightly, tested and seems okay.
- Only sticky on desktop displays.
- Tables of contents which are longer than the page height get their own y-axis scroll.

Added a jump to top link for mobile displays

4.20.2

Neither.

## Checklist

- [x] Checked on Safari
- [x] Checked on Firefox
- [x] Checked on Chrome

### DIFF
--- a/assets/scss/gds-design-system-settings.scss
+++ b/assets/scss/gds-design-system-settings.scss
@@ -16,8 +16,13 @@ $govuk-font-family: 'PT Sans', sans-serif;
 
 
   body {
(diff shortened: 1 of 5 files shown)


RobLoweMOJ (approved):
Discussed. - Fix is in the PR for scroll spy

PR #419, opened 2025-02-20
Scrollspy

MalcolmVonMoJ:

Uses JS to highlight where the user is on the page on the Table of Contents:
- Adds a class to mark the relevant item in the ToC.  
- If at the bottom, always marks bottom item.

Styles current item:
- disables link
- uses text colour
- corrects overflow-y style for ToC

Adds a print this page button
- button hidden on print and if no JS

4.21.8

Dev

## Checklist

- [x] Checked on Safari
- [x] Checked on Firefox
- [x] Checked on Chrome

### DIFF
--- a/assets/js/navigation.js
+++ b/assets/js/navigation.js
@@ -309,3 +309,47 @@ jQuery("#menu-menu-top-menu li.menu-item-has-children > ul.sub-menu").ready(func
 		if (e.keyCode == "13") $(this).click();
 	});
 });
+
+/**
+ *
+ * Table of contents "scrollspy", highlight current scroll location
+ *
+ */
+
+( function() {
+	indicateCurrentLocation();
+})();
+
+document.addEventListener('scroll', function() {
+	indicateCurrentLocation();
+}, false);
+
+function indicateCurrentLocation(){
+	if (!document.querySelector("#toc")) return;
+	let toc = document.querySelector("#table-of-contents");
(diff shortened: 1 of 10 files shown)


RobLoweMOJ:
Found an edge case. Where some links get skipped.

assets/js/navigation.js
```diff
@@ -309,3 +309,47 @@ jQuery("#menu-menu-top-menu li.menu-item-has-children > ul.sub-menu").ready(func
...
+		document.body.scrollHeight,
+		document.body.offsetHeight,
+		document.documentElement.clientHeight,
+		document.documentElement.scrollHeight,
+		document.documentElement.offsetHeight
+	);
+	const windowHeight = window.innerHeight;
+	const maxScroll = documentHeight - windowHeight;
+
+	if (window.scrollY >= maxScroll) i = sectionHeadings.length - 1;
```
MalcolmVonMoJ, line 348: This bit of code is what's causing the skip sections - by design.  If you're at the bottom of the page, the last item is always "current".  

This means that if someone clicks on the last item and it is too far down the page to be at the top - it is still highlighted.  

Of course, if there are lots of small sections, then some will be missed.  I thought this approach would deal with the most likely scenario.

MalcolmVonMoJ: I have asked Lewis about this behaviour to confirm it is okay.

RobLoweMOJ (approved):
LGTM

PR #422, opened 2025-02-21
Footer crest fix

MalcolmVonMoJ:

Fixes the colours for the royal crest on GDS styled sites

4.20.4

Neither.

## Checklist

- [ ] Checked on Safari
- [ ] Checked on Firefox
- [x] Checked on Chrome

### DIFF
--- a/style.css
+++ b/style.css
@@ -1,7 +1,7 @@
 /*
 Theme Name: Hale
 Text Domain: hale
-Version: 4.20.3
+Version: 4.20.4
 Domain Path: /languages
 Description: Theme for Ministry of Justice websites.
 Author: Ministry of Justice


assets/scss/brandings.scss
```diff
@@ -1193,5 +1186,13 @@ body.hale-colours-variable {
 			}
 		}
 	}
+}
 
+.govuk-footer {
```
RobLoweMOJ, line 1191: WHy are we moving this outside the .hale-colours-variable class?

MalcolmVonMoJ: This is to make the default colour work - the variable colours were working as intended, but the default wasn't being picked up.  This should always have been applied to all sites, not just those with custom colours.  

I noticed this on Adam's new site which is GDS styled, but it also appears on the Dashboard and elsewhere.

RobLoweMOJ: Sounds good

PR #425, opened 2025-02-24
New heading option for CPTs

MalcolmVonMoJ:

Added an option to move the heading to the very top of the page and give it a shade for CPTs

4.20.5

Neither.

## Checklist

- [ ] Checked on Safari
- [ ] Checked on Firefox
- [x] Checked on Chrome

### DIFF
--- a/assets/scss/brandings.scss
+++ b/assets/scss/brandings.scss
@@ -758,6 +758,16 @@
 		}
 	}
 
+	//Shaded headings (full-width headings)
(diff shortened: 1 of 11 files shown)


RobLoweMOJ:
COuld pass a param into the templates for if the header if the h1 should display. Would reduce repeat calls to DB. Also this should work with document upload switched on

RobLoweMOJ (approved):
LGTM

PR #426, opened 2025-02-24
Revision date

RobLoweMOJ:

Neither.

## Checklist

- [ ] Checked on Safari
- [ ] Checked on Firefox
- [ ] Checked on Chrome


brown-a2 (approved):

brown-a2:
I assume it wasn't in scope but would we also want revision date to show up on the single page?

MalcolmVonMoJ:
> I assume it wasn't in scope but would we also want revision date to show up on the single page?

There is a date in the pattern for the new full-width header.  My PR for that currently uses the updated date - when this is merged I can add the revision date to that logic, probably as a new PR.

brown-a2 (approved):

brown-a2:
Only small thing, in the ACF label you have it as Restrict blocks...but maybe it should just be 'Blocks' bescuse the restricting it really happening below in the include/exclude field. Just a suggestion, either way looks great.

PR #427, opened 2025-02-25
Cpt restirct blocks

RobLoweMOJ:

Allows you to restrict blocks to on a CPT

Neither.

## Checklist

- [ ] Checked on Safari
- [ ] Checked on Firefox
- [ ] Checked on Chrome


brown-a2 (approved):

brown-a2:
Only small thing, in the ACF label you have it as Restrict blocks...but maybe it should just be 'Blocks' bescuse the restricting it really happening below in the include/exclude field. Just a suggestion, either way looks great.

PR #431, opened 2025-02-27
Remove callback function as editors dont have access to colours

RobLoweMOJ:

Fix fatal error  for site editors - Remove callback function as editors dont have access to colours

TBC

Neither.

## Checklist

- [ ] Checked on Safari
- [ ] Checked on Firefox
- [ ] Checked on Chrome


MalcolmVonMoJ (approved):
😭

PR #436, opened 2025-03-12
Remove latest news from authorised blocks

MalcolmVonMoJ:

This PR prevents new Latest News blocks from being added, the Item List block should be used instead.

4.21.7

Dev

## Checklist

- [ ] Checked on Safari
- [ ] Checked on Firefox
- [x] Checked on Chrome


RobLoweMOJ (approved):
LGTM

PR #441, opened 2025-03-31
Fix SVG Upload - was not working in customizer

RobLoweMOJ:

Fix SVG Upload - was not working in customizer

Neither.

## Checklist

- [ ] Checked on Safari
- [ ] Checked on Firefox
- [ ] Checked on Chrome


brown-a2 (approved):
I think I'd tend to expect it in the uploads.php file but I understand why you put it in here.

PR #442, opened 2025-04-03
Ticket taylor

MalcolmVonMoJ:

Completely re-does the title shading code as the previous approach interferes with the Ticket Taylor code.  

The new approach uses box shadows - which is a bit more elegant.  As we cannot z-index box shadows, it might cover up other things - but I don't think we ever had a situation where the shading goes beneath something else.  

This only affects normal page and search page titles.

4.21.13

Neither.

## Checklist

- [x] Checked on Safari
- [ ] Checked on Firefox
- [x] Checked on Chrome

### DIFF
--- a/assets/scss/brandings.scss
+++ b/assets/scss/brandings.scss
@@ -1070,32 +1070,8 @@ body.hale-colours-customised {
 
 	.hale-search-header,
 	.page-header-section {
-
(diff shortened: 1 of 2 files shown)


RobLoweMOJ (approved):
LGTM

PR #447, opened 2025-04-22
Topnav errors & mobile errors

MalcolmVonMoJ:

3 bugs are fixed here:
- On situations with no breadcrumbs, the last item of the main navigation (mobile view only) spills over the bottom of the menu.  This is fixed here by applying margin and padding as needed to items which follow the header if they are divisions.  
- In mobile view, there is some horizontal scrolling happening.  This has been fixed by applying an `overflow-x:clip` property to the `html` tag.  
- The latest Chrome update seems to have removed support for `a:focus:before` in CSS.  It is still supported in Firefox.  I am working on the assumption that this is a bug which will soon be fixed.  In the meantime, I have made our CSS more specific to ensure that hover styles are not applied when focussed (`:hover:focus` isn't working, we don't want `:hover` to be the style).  Luckily `:hover:not(:focus)` is still working.  If this isn't fixed in Chrome soon, we might have to revisit this.  

4.21.15

Neither.

## Checklist

- [x] Checked on Safari
- [x] Checked on Firefox
- [x] Checked on Chrome

### DIFF
--- a/assets/scss/brandings.scss
+++ b/assets/scss/brandings.scss
@@ -412,7 +412,7 @@
 							@include hale-heading-link-underline-restrictor($colour: --header-link-focus-highlight);
 						}
 
-						&:hover {
+						&:hover:not(:focus) {
 							color: var(--header-link-hover);
 							border-bottom-color: var(--header-link-hover-border);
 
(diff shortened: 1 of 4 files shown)


RobLoweMOJ (approved):
LGTM

PR #453, opened 2025-05-22
Remove package lock as theme is compiled on build

brown-a2:

Removes Package Lock file from repo and adds it to gitignore.

We do not need this file as it is generated when we compile and build the Docker image. 

Demo

## Checklist

- [ ] Checked on Safari
- [ ] Checked on Firefox
- [ ] Checked on Chrome


RobLoweMOJ (approved):
LGTM

PR #457, opened 2025-06-18
List margin

MalcolmVonMoJ:

Brings through the GDS list styles for list blocks (previously worked but I guess WordPress changed this block to have a class which broke it).  

4.24.2

(what a pleasing number)

Neither.

## Checklist

- [ ] Checked on Safari
- [ ] Checked on Firefox
- [x] Checked on Chrome


RobLoweMOJ (approved):
LGTM

PR #462, opened 2025-08-01
Add autocomplete filter to listing page - part one

brown-a2:
This PR adds an autocomplete filter by taxonomy to the listing template. This is controlled via ACF. 

PR breakdown:
- Adds a toggle in the backend of ACF on a taxonomy that switches autocomplete on or off.
- When toggled on, any listing template filter using that taxonomy will switch to use an autocomplete filter.
- Uses same Gov UK css
- Introduces a new JS file just for using autocomplete on the listing template. 
- Uses same vars passed into the existing filter group.
- Minor formatting or PSR12

Still to do (part 2)
- Tidy up the look of the autocomplete field - arrow too big.
- While some work has been done on subtopics already, there is still some work to do to get it fully working.

### DIFF
--- a/assets/js/autocomplete-filter.js
+++ b/assets/js/autocomplete-filter.js
@@ -1,24 +1,206 @@
-/* autocomplete filter for listing page */
+/* Simplified autocomplete filter for listing page */
+var taxonomies_with_terms = {};
 
+// Check if the object exists and extract terms
+if (typeof listing_page_object !== 'undefined') {
+    console.log('listing_page_object:', listing_page_object);
+    
+    // Store taxonomy terms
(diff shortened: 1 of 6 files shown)


RobLoweMOJ:
You are using a mix of localisation scripts and document.querySelectorAll to load data into the js.. the code works i am just wondering if this should be consistent. happy to discuss.

brown-a2:
> You are using a mix of localisation scripts and document.querySelectorAll to load data into the js.. the code works i am just wondering if this should be consistent. happy to discuss.

Sure happy to discuss. ... my rational is that I've tried to keep a separation of concerns where the global data comes via PHP localisation while the instance-specific data is managed via the dom query selector look ups.

RobLoweMOJ (approved):
LGTM

PR #465, opened 2025-09-18
Add Column Block to allow list

RobLoweMOJ:

Add Column Block to allow list

Neither.

## Checklist

- [ ] Checked on Safari
- [ ] Checked on Firefox
- [ ] Checked on Chrome


brown-a2 (approved):

brown-a2:
A quick look and it seems we are missing quite a few core blocks from this list. We might need to revisit this at somepoint especially if we are starting to use more patterns.

PR #509, opened 2026-07-15
Listing images

MalcolmVonMoJ:

Adds a new option to include the featured image in the listing page.
Adds a new option to hide the post type name and the updated date from the single view with full-width heading.

4.34.5

Demo

## Checklist

- [ ] Checked on Safari
- [ ] Checked on Firefox
- [x] Checked on Chrome


functions.php
```diff
@@ -183,6 +183,13 @@ function hale_content_width()
  * Enqueue scripts and styles.
  */
 
+add_action( 'admin_enqueue_scripts', function() {
+    wp_enqueue_style(
+        'hale-admin',
+        hale_mix_asset('/css/admin.min.css')
```
RobLoweMOJ, line 199: Can we be mroe specific on where this loads.

RobLoweMOJ (approved):
LGTM - though css could be restricted further

PR #520, opened 2026-07-29
Update npm deps

earthlingdavey:

Updates NPM packages.
Includes some minor changes to scss to avoid deprecated syntax, css output before & after is the same.
Turns off noisy deprecation warnings when running `npm run production` etc..

4.35.1

Demo

## Checklist

- [ ] Checked on Safari
- [ ] Checked on Firefox
- [x] Checked on Chrome

## Notes

The updated in this PR are quick wins, doesn't include updates to `@ministryofjustice/frontend` or `govuk-frontend`. And, doesn't include migrating `@import` rules to `@use`/`@forward`.


MalcolmVonMoJ (approved):
LGTM

PR #521, opened 2026-08-04
Remove Heading block variants being unregistered and disable fit text

RobLoweMOJ:

Remove Heading block variants being unregistered and disable fit text

In Wordpress 7.0.2 it uses variants of the heading block for heading sizes e.g. h2 h3 h4

We originally unregistered variants as to prevent use of the stretchy heading and paragraph blocks but this has now moved to a fit text setting. 

4.35.1

Neither.

## Checklist

- [x] Checked on Safari
- [x] Checked on Firefox
- [x] Checked on Chrome

### DIFF
--- a/assets/js/restrict-embed-blocks.js
+++ b/assets/js/restrict-embed-blocks.js
@@ -50,12 +50,4 @@ wp.domReady(function () {
     }
   });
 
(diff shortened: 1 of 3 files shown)


earthlingdavey (approved):
LGTM

